PR #7, opened 2020-02-25
Added: formatting tools and formatting to ilapfuncs.py

agamdua:
Added two formatting tools that I use in most of my projects. These tools format python source code to industry standards so there isn't a lot of bike-shedding on programming style.

I have restricted formatting to one file, the one I currently care about. Over time we will adopt the other files.

https://black.readthedocs.io/en/stable/
https://github.com/miki725/importanize

This can be run by calling

```
$ make format
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,3 @@
+format:
+	black ilapfuncs.py
+	importanize ilapfuncs.py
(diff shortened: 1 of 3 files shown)


abrignoni:
I will start reading and learning these now.

PR #16, opened 2020-03-01
Add files via upload

abrignoni:
Started os.path.join support.
Modified KnowledgeC to support parsing per full iOS version numbers.

### DIFF
--- a/ilapfuncs.py
+++ b/ilapfuncs.py
@@ -1,3 +1,4 @@
+#ileapp
 import base64
 import codecs
 import csv
@@ -14,6 +15,7 @@ import sqlite3
 import sys
 import textwrap
 from time import process_time
+from packaging import version
 
 from bs4 import BeautifulSoup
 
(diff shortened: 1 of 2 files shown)


ilapfuncs.py
```diff
@@ -65,8 +67,8 @@ def datark(filefound):
     logfunc(f"Data_ark.plist function executing")
     deviceinfo()
     try:
-        os.makedirs(reportfolderbase + "Data_Ark/")
-        with open(reportfolderbase + "Data_Ark/Data Ark.html", "w") as f:
+        os.makedirs(os.path.join(reportfolderbase, "Data_Ark/"))
+        with open(os.path.join(reportfolderbase, "Data_Ark/Data Ark.html"), "w") as f:
```
agamdua: When using `os.path.join` you want to pass all paths as an arg, otherwise you're still hardcoding the separator.

Another option is to use `os.sep` and string formatting, but since we have chosen the former in ALEAPP just makes sense to follow that convention.

PR #18, opened 2020-03-02
Massively simplified dbbuffer function

agamdua:


agamdua:
Just noticed something I want to improve, will re-open in a bit

abrignoni:
10-4

contrib/aggregated_dictionary/main.py
```diff
@@ -76,3 +84,22 @@ def aggdict(filefound):
...
+
+    db_path = filefound[0]
+    category = "DBBuffer"
+
+    _create_aggregate_dict_dir()
+
+    with open(db_path, "r") as rfp:
+        rows = [line.split() for line in rfp.readlines()]
+
+    report_table_columns = ["Value"] * 4
```
agamdua: @abrignoni - I think we should create better table headers for this module, unless it is intentionally all the same. I saw this in the previous code and maintained it, but at least in the iOS image I am using the values are pretty distinct. Any suggestions? 

This change I am suggesting can be separate from this particular PR, and would require some of your domain expertise.

abrignoni: The data in the dbbuffer text file seems to be a buffer for the ADDataStore.sqlitedb. Some of the columns are obvious (like the first column being a timestamp <shame> which I should have converted to human readable UTC but I haven't  </shame>) but the others in not super sure about. It looks just like scalars but with an added column. Decided to parse it for completeness regarding  items not yet pushed into the DB. Any thoughts on better table headers will be well received.

agamdua: I don't really have a good idea unfortunately, so if there isn't an obvious candidate I would say leave it as is.

Issue #27, opened 2020-03-10
iLEAPP GUI Error in Latest Commit


chipriley:
I receiving the following error when attempting to analyze a GrayKey full filesystem extraction with the latest commits.  I ran an analysis a couple of days ago with a previous version and it completed without error.  Here's the traceback:

```python traceback
Traceback (most recent call last):
  File "ileappGUI.py", line 115, in <module>
    extract_and_process(pathto, extracttype, tosearch, log, gui_window=window)
  File "[redacted]\extraction.py", line 71, in extract_and_process
    process_file_found(filefound, key, val, log, gui_window)
  File "[redacted]\extraction.py", line 127, in process_file_found
    globals()[key](filefound)
  File "[redacted]\ilapfuncs.py", line 1124, in mib
    file = open(filename, "r", encoding="utf8")
FileNotFoundError: [Errno 2] No such file or directory: '\\private\\var\\installd\\Library\\Logs\\MobileInstallation\\moile_installation.log.0'
```
Not sure if something changed but it shuts down the GUI and stops parsing.

abrignoni:
Fixing now

abrignoni:
Issue at os.path.join() at search_file.py.
Fixed applied. Will leave open as marker to follow up with a more efficient solution soon.

chipriley:
That fixed it.  

On a possibly related note (but it might need to be a separate issue), I just tried the latest committed version against a GrayKey BFU extraction and got the following traceback error:

``` python traceback
Traceback (most recent call last):
  File "ileappGUI.py", line 115, in <module>
    extract_and_process(pathto, extracttype, tosearch, log, gui_window=window)
  File "[redacted]\iLEAPP\extraction.py", line 71, in extract_and_process
    process_file_found(filefound, key, val, log, gui_window)
  File "[redacted]\iLEAPP\extraction.py", line 127, in process_file_found
    globals()[key](filefound)
  File "[redacted]\iLEAPP\ilapfuncs.py", line 294, in knowledgec
    """
sqlite3.OperationalError: no such table: ZSTRUCTUREDMETADATA
```
I'm guessing that the absence of something in the BFU extraction is causing it to crash.

abrignoni:
Fixing now.

abrignoni:
Put some try except clauses around the KnowledgeC artifacts so the program won't crash.
There might be more needed. Thanks for testing on a BFU extraction @chipriley. I will get one tomorrow so i can do more testing and fixes.

[...]

PR #29, opened 2020-03-11
Added: os.path.join back for zip archives

agamdua:
Previously, os.path.join would discard the base path on some archives
if the path they returned were absolute paths.

This workaround allows for such paths and keeps it platform-agnostic.

Please test, although I have run a test on the latest zip I have.


agamdua:
This should close #27

PR #44, opened 2020-05-06
Updated requirements.txt to more accurately reflect what is required :)

edward-greybeard:
Hey I was perusing this fancy new branch and realised you appear to have a bunch of now-unused requirements in your requirements.txt. So I removed them! Still not sure about pathlib2 though.


abrignoni:
Thanks for the necessary cleanup. 💖 I think I use pathlib for the notifications artifact. Will check later on today

PR #46, opened 2020-05-10
update to notifications

ydkhatri:
Uses my deserializer for NSKeyedArchives. Fixed parsing and gets bundle names (for bundle ids) from the UserNotificationsServer/Library.plist db. Now getting lots of notifications on josh's image.

This isn't complete yet. I'm not extracting embedded plists. Need to add that functionality back. But the code is much simpler now.


abrignoni:
This looks fantastic!!!! Thank you so much. Merged with great pleasure.

PR #53, opened 2020-06-22
Add Ookla Speed Test App Parsing

mastenp:
Added support for the Ookla Speedtest Application including data related to each time an internet "speed test" was conducted.  Data includes Date/Time, External IP, Internal IP, Cellular Carrier (if test done over cellular connection),Device Model, ISP, WAN type, Wifi SSID (if test done over wifi connection), Lat/Long, Lat/Long Accuracy.  Sample of the report is below in the image. 

Screen Shot 2020-06-22 at 3 11 38 PM


mastenp:
Just FYI 
-The Delete infoList.py is for another artifact that I didn't mean to bring in just yet, but Github Desktop grabbed it.
-The Delete ooklaSpeedtest.py was the script that I couldn't get working, so the working one is ooklaSpeetestData.py

mastenp:
Added WAL file capability and updated the search for the database for Ookla Speed Test.

ydkhatri:
Can you share a sample database (**speedtest.sqlite**) to test this with?

ydkhatri:
Code looks good, but please remove the unused imports, like ccl_bplist, plistlib (and others) which aren't used.

mastenp:
Removed unused Imports as requested.

PR #54, opened 2020-06-24
Filesystem cache update

edward-greybeard:
Updated the FileSeekerDir.build_files_list() to use os.scandir(). Tested on hickman's iOS 13-3-1 full FS extraction and it went from 60-80 secs to build the list to 20 secs.


ydkhatri:
Indeed, this is another 50-60% faster now! On my files, processing goes from 27 seconds to 12 seconds. Thanks for the fix.

PR #57, opened 2020-07-21
Added support to compile to an executable(s) using pyinstaller

markmckinnon:
Added pyinstaller requirement to requirements.txt file.
Added how to compile CLI and GUI scripts using pyinstaller
Create spec files used to compule scripts to single file executables.


abrignoni:
This is fantastic!
Thank you so much. 💖

PR #67, opened 2020-08-08
remove os.getcwd() in location determination code

ydkhatri:
Please test with the exe version.

### DIFF
--- a/ileapp.py
+++ b/ileapp.py
@@ -33,6 +33,12 @@ def main():
         print('INPUT file/folder does not exist! Run the program again.')
         return  
 
+    # ios file system extractions contain paths > 260 char, which causes problems
+    # This fixes the problem by prefixing \\?\ on each windows path.
+    if is_platform_windows():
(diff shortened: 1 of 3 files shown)


abrignoni:
Downloaded the fork and created the binaries.

The cmd line version produced errors in the format of “Could not write file
to file system, path was “ for the snapshot artifact. The report showed
only 18 snapshots instead of the expected 290 + snapshots.

The GUI version showed 297 snapshots and no errors in the log.

The notifications looked fine on both. I used Hickman’s image.

ydkhatri:
> The cmd line version produced errors in the format of “Could not write file to file system, path was “ for the snapshot artifact. The report showed only 18 snapshots instead of the expected 290 + snapshots.

Hmm, then that is probably not related to this change, but I will take a look at the cmd version and get back to you on it.

ydkhatri:
This had to do with long file paths, now fixed..

abrignoni:
Release has been updated.
Windows 10 64-bit.

PR #71, opened 2020-08-27
Geolocation cache

flamusdiu:
This seems to be a better pull request. Not sure what was going on before.


abrignoni:
Worked great on my sample images.

PR #74, opened 2020-09-06
Add decoding for tile location names and label names

ydkhatri:
* Add decoding for tile location names and label names. This will identify the streams from the blob, decompress them, then decode.. There are a few binary streams too, with unknown data.. so there is scope for more research here.
* Changed tabs to spaces
* Hex display for Keys A B C D, and tileset - In hex, you can see some sort of pattern in the data, needs more research to decode
* Removed hexdump and etag (from html output) as it is taking up unnecessary space on the page and does not really help you analyze or interpret anything from there. Besides most of the info in the blob is compressed. If someone needs to look at the hex, they should go to the db directly.

Tested with ios 12 and 13 images. But should probably be tested more..

Output looks like this:
geotiles-changes


abrignoni:
Tested. Worked as expected on iOS 12 & 13. None of my iOS 11 images had data for the artifacts.

PR #78, opened 2020-09-12
Add build pipeline for iLEAPP executables

forensicmike:


ydkhatri:
Mike 

Thanks for submitting this. I did merge your workflow. I don't have any experience with github workflows, so unsure how to debug this. It seems to fail with this error:
```Validation Failed: {"resource":"Release","code":"already_exists","field":"tag_name"}```

I am guessing that it wants to create a new release and won't update an existing one. But I am triggering this by creating a new release to begin with..

PR #85, opened 2020-09-22
App Group Listing

abrignoni:
Artifact parses each .com.apple.mobile_container_manager.metadata.plist file and extracts the corresponding bundle ID & app group GUID.

It is of note that a deleted app can leave behind the app group directory with possibly relevant data. The plist identifies the app that was related to it even after the app has been deleted.


abrignoni:
Adding to the Installed Apps category. A note on the report indicates the list can contain bundle ID for apps that have been deleted.

PR #100, opened 2020-10-09
Licensing

ydkhatri:
Take a look, I have only attributed one query, please do so in the other files as well. The license(s) don't need to be on wiki, you can remove from there. Just add the footnote of Acknowledgements as in README.md


abrignoni:
Will do. Will start making changes tomorrow.

PR #104, opened 2020-10-12
Update photosMetadata.py

abrignoni:
Edited to include more data to include the extracted postal address from the original location nskeyedarchived bplist blob.
Credit to https://twitter.com/Scott_Kjr for the queries.


abrignoni:
Also two different queries. One each for iOS13 & iOS14

PR #110, opened 2020-10-24
Fix bug in wifi script, merge duplicative scripts, add support for networks scanned with private mac

### LINKED ISSUE #88: Need to consolidate 2 wifi plugins
There is 
`            'appleWifiPlist':('Wireless Networks', '**/SystemConfiguration/com.apple.wifi.plist'),`
and
`            'wifi':('Wifi Connections', '**/com.apple.wifi.plist'),`

image

Both provide similar data, need to be merged.

threeplanetssoftware:
This PR was generated from seeing errors in the `scripts/artifacts/appleWifiPlist.py` script when the object did not contain a 'List of known networks' key. In fixing that, I also combined the information which was in `scripts/artifacts/wifi.py` into this script, per #88. I added support to `scripts/artifacts/appleWifiPlist.py` for a newer object wherein Apple records the networks the phone scans while using a "private" MAC address. Finally, I added three files to the file list for this script as the backup files were present on one of my test devices when the expected file was not. 

One question, `scripts/artifacts/appleWifiPlist.py` would shove the entire object into the output array. This leads to some very tall rows in Hickman test images, I'd personally lean towards ripping that out, but wanted to check before doing so. 

Fixes #88.


abrignoni:
Tested it and it worked great. Thanks for the added capability. Going to merge as is now since it is a really useful artifact.
If you could, at your convenience, rip out the data and put them after the ones already on the report (since they are more relevant) that would be awesome. 

Thanks for this great addition. Those wifi artifacts you added would make a for a great informative blogpost. Just saying. :-D

threeplanetssoftware:
By "rip out the data and put them after the ones already on the report" do you mean take that column and make it is own table later on?

abrignoni:
I assumed you were going to take the values on the long column and spread them out on the report so they would look like the rest of the report entries. Maybe I misunderstood.

threeplanetssoftware:
It looked like someone was just trying to throw the entire object in there. Pulling every key out would make for a very long table. I doubt you'd want to hard code that many column names and would suggest if you want literally everything displayed, to do it programatically. I, personally, think it more important to display the most worthwhile artifacts and let the examiner dive deeper into the others if they need to and would be in favor of just dropping that column entirely.

abrignoni:
Good points. Agreed.

PR #112, opened 2020-10-24
Change Longitude to Logitude to match scripts/artifacts/knowClocation.py

### LINKED ISSUE #111: Longitude not consistently spelled, knowCall throws errors
Running the current version of iLEAPP against test data which has knowledgeC information in it leads to this error:

``` python
Incepted bplist extractions in KnowledgeC.db completed
Reading KnowledgeC artifact had errors!
Error was 'Longitude'
  File "/home/notta/data/programming/python/iLEAPP/scripts/ilap_artifacts.py", line 406, in process_artifact
    method(files_found, report_folder, seeker)
  File "/home/notta/data/programming/python/iLEAPP/scripts/artifacts/knowCall.py", line 1443, in get_knowCall
    kmlgen(report_folder, kmlactivity, data_list, data_headers)
  File "/home/notta/data/programming/python/iLEAPP/scripts/ilapfuncs.py", line 194, in kmlgen
    lon = modifiedDict['Longitude']
KeyError: 'Longitude'
```

It appears the root cause is in the following line wherein the data headers specify 'Logitude' as the field name, not 'Longitude'. This leads to the `keyError`. My preferred fix would be to fix the spelling in the below line back to 'Longitude' as that is the name used in the original database, but the choice is yours. 

https://github.com/abrignoni/iLEAPP/blob/68c08a19d2eac05759a5d891fcd2b1194afe2870/scripts/artifacts/knowClocation.py#L91

Because I believe the exported TSV already had ingest artifacts written for it for other tools, I've made a change to just check for 'Logitude', instead of changing the data headers to 'Longitude'. PR with the fix is incoming.

threeplanetssoftware:
This PR normalizes the spelling of Longitude between `scripts/artifacts/knowClocation.py` and `scripts/ilapfuncs.py` to be "Logitude". This matches the data headers when generated.

Fixes #111.

### DIFF
--- a/scripts/ilapfuncs.py
+++ b/scripts/ilapfuncs.py
@@ -191,7 +191,7 @@ def kmlgen(report_folder, kmlactivity, data_list, data_headers):
     while a < length:
         modifiedDict = dict(zip(data_headers, data_list[a]))
         times = modifiedDict['Timestamp']
-        lon = modifiedDict['Longitude']
+        lon = modifiedDict['Logitude']
         lat = modifiedDict['Latitude']
         if lat:
             pnt = kml.newpoint()


abrignoni:
If we change the header in the function to Logitude al the other properly spelled headers will not work when the function is called. I will change the misspelled header in knowClocation.py and that should fix it.

Will check to make sure there are no other artifacts with misspelled headers.

abrignoni:
Changed the header. Hopefully I didn't fat fingered the header values anywhere else. Thank you so much. :-D

Issue #70, opened 2020-08-27
iLEAPP fails if output dir is on a network share


jarlethorsen:
When setting the output directory to a network share (and using an archive as input) I get a lot of errors like this:

> Safari Browser [safariHistory] artifact executing
> Reading Safari Browser artifact had errors!
> Error was database is locked
> Exception Traceback: Traceback (most recent call last):
>   File "/home/user/Sources/iLEAPP/scripts/ilap_artifacts.py", line 359, in process_artifact
>     method(files_found, report_folder, seeker)
>   File "/home/user/Sources/iLEAPP/scripts/artifacts/safariHistory.py", line 58, in get_safariHistory
>     timeline(report_folder, tlactivity, data_list)
>   File "/home/user/Sources/iLEAPP/scripts/ilapfuncs.py", line 134, in timeline
>     cursor.execute('''PRAGMA journal_mode = WAL''')
> sqlite3.OperationalError: database is locked

Sqlite has a known issue with database locking when the database is located on a network share (I have also seen this in my own projects)

Maybe you can catch the "_sqlite3.OperationalError: database is locked_" exception and inform the user that this may be a result of having output path on a network share,

or

use _tempfile.TemporaryDirectory()_ to create the temporary directory. (It even comes with a context-manager that takes care of cleaning up afterwards). This way the temp dir will be created in the default system location for the user, but the result may still be created wherever the user chooses?

ydkhatri:
Can you provide more details? Is the input a folder/zip/tar? Is the output also on a network drive? Are you accessing via drive letter or //server/folder path? 
Is your network drive read-only?

jarlethorsen:
The problem here is related to running _cursor.execute()_ on a sqlite3 database that happens be located on a network share.

A google search turns up a dozen similar issues in other projects, related to storing the database on a network share, like this one: https://stackoverflow.com/questions/57248340/how-do-i-unlock-a-database

In the case of iLEAPP this is directly related to where the temporary directory is created.

If the input is a folder or gz does not make any difference, (this makes sense, as I would imagine you copy the db to the temp-dir before you open it, no matter the type of input, to avoid making changes to the original db?)

Personally I'm using iLEAPP on a linux system where the network shares are mounted cifs shares. (None of the network shares are read only).

For example I can process '**/Safari/History.db' just fine, no matter if the source is a folder or a gz, or the source is on a network share or local disk, the only thing that will make it fail is setting the output to anywhere on a network share.

[...]

PR #117, opened 2020-11-04
modified file copy to handle NFS share mount permissions

xperylab:
When using iLEAPP on nfs mounts, it can not modify permissions (set by nfs server). shutil.copy2 (or shutil.copy) is also copying permissions from files in git repo. 
It makes sense to inherit of local output directory and it make it work on nfs mounts :)

Also deleted for same reason some chmod in bluetooth scripts. No sure what is their use...

Do not solve issue #70 as it concern samba, but same kind of problem.

NB : My weird error handling would be solved by python 3.8 and option dirs_exist_ok in shutil.copytree.


abrignoni:
Thank you so much!!!!

PR #118, opened 2020-11-06
Photodata ios12

xperylab:
Added support for photo metadata artefact for iOS 12 (tested on iOS 12.1.4)
Added thumbnails to the report table (only for ios12 for now - will test on other when I process dumps)


abrignoni:
This is a cool functionality. 
Some thing I noticed:
1. The data_header and the data_list dont match. It seems the error starts at the Moment Approximate Latitude field.

Screen Shot 2020-11-06 at 9 53 59 AM

2. The thumbnails didn't show up on iOS version: 12.4.4.

Screen Shot 2020-11-06 at 9 58 10 AM

Thank you so much for this. This is amazing stuff. Do you have a Twitter account?

xperylab:
thanks for testing !
1.  oops, I deleted a line in the query... nice catch !
2. I put more generic paths and added a thumbnail creation in case the thumbnail is not found.
Do not work on videos though...
Can you give me the path of thumbnails on 12.4.4 (like PhotoData/Thumbnails/....) ?

I created my pro Twitter account and linked it to my profile (@xperylab)

abrignoni:
Just looked. The path should be the same for all iOS versions.
Tried it again after you latest changes and it works for iOS 12.4.4.
I bet if you take out the version restriction it will work for iOS 13.

Check out Josh Hickman images here for testing purposes:
https://thebinaryhick.blog/2020/04/16/ios-13-images-images-now-available/

xperylab:
Yes thumbnail may also work for ios13. However the query is not exactly the same so I have to copy/paste, adapt and test with Josh images :)

abrignoni:
You are absolutely correct. Looking forward to the updated script. :-)

xperylab:
refactored a bit to include the thumbnail function in ilapfuncs instead of copy/paste. Tested on 12.1.4, 13.3.1 and 13.4.1 with Josh images.

PR #119, opened 2020-11-06
Add location vizualisation map

xperylab:
I have been working on a way to visualise all locations at once (except photos). 
Takes files in KML export and plot on an interactive map (with a time slider). Only tested for iOS 12.1.4. 

Routine D ZRTXX are plotted as line because it seems to make more sense. All others are plotted as circles (wi-fi/cell). 

Using circles (instead of markers) because they are lighter and browser can remain fluid with a lot of them.

sample result on my dataset:
image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,6 @@ blackboxprotobuf
 nska-deserialize
 biplist
 simplekml
+folium
+pandas
+geopandas
--- a/scripts/location_map_constants.py
+++ b/scripts/location_map_constants.py
@@ -1,0 +1,106 @@
+# coding: utf-8
+iLEAPP_KMLs = "_KML Exports"
+
+defaultShadowUrl = 'https://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.7/images/marker-shadow.png'
+
+defaultIconUrl = 'https://raw.githubusercontent.com/pointhi/leaflet-color-markers/master/img/marker-icon-2x-{}.png'
+
+colors = ['red','green','blue','orange','violet','yellow', 'sienna', 'indigo','purple','cyan','darkgray']
+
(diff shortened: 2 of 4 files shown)


abrignoni:
This functionality is INSANE!!
It works on iOS 12.4.4 and on iOS 13.4.1

Important fix: If the user does not parse an artifact that generates KML files the locations map function crashes the program. Needs to check if there are KML files available before processing.

Question: How will the locations map functionality behave if there is no internet?

Only one small detail. The page header for the artifact changes. Im using Chrome.
Artifact header:
Screen Shot 2020-11-06 at 10 26 14 AM

Artifact header from Location map:
Screen Shot 2020-11-06 at 10 24 37 AM

Really looking forward to merging this. It is a fantastic addition. I can't thank you enough.

xperylab:
Thanks !! Happy that you find this useful !

> I fixed the map generation to return if no KML export

> If no internet connection, it won't work. It is possible to download the JS scripts but for the tiles it does not make so much sense, so... 
It is still possible to pass an option to the map to use an offline server for tiles but it is huge...

> Page headers also change on Safari.... Probably a conflict with the CSS because it is raw writing an html into another html.. 
will try to find what is wrong but not top priority IMHO :)

abrignoni:
> Thanks !! Happy that you find this useful !
> 
> > I fixed the map generation to return if no KML export
> Works great!
> > If no internet connection, it won't work. It is possible to download the JS scripts but for the tiles it does not make so much sense, so...
> > It is still possible to pass an option to the map to use an offline server for tiles but it is huge...
> Not needed. Was just wondering.
> > Page headers also change on Safari.... Probably a conflict with the CSS because it is raw writing an html into another html..
> > will try to find what is wrong but not top priority IMHO :)
> I agree. Not a priority.

PR #139, opened 2020-11-17
Scroll bar hold position feature

ydkhatri:
Also fixed the report source path display (for the right kind of slash) on windows.


abrignoni:
The feature works in ALEAPP but not on iLEAPP.
I have macOS and opening the report on Google and Safari.

ydkhatri:
I believe it is now fixed, late last night. Try it now.

PR #142, opened 2020-11-19
Location map2

xperylab:
Here is the PR to re enable locations map. 
I also updated version_info.py :)


JamesHabben:
we are moving away from using the static html output, and there are conflicts in here that dont align with current architecture. create a new PR if there is still valuable code to contribute and we can review.

PR #148, opened 2020-11-22
Faster walStrings

ydkhatri:
Replaced the code to use regex. This is much faster now. Also added deduplication, so strings don't show up multiple times.
The speed improvement is big. On the one I am testing on, it went from 40 seconds down to 7 seconds (ascii char matching) or 8 seconds (unicode char matching).

I'm not merging this yet because there are some issues to be considered. The old code was only getting ASCII chars, as it was filtering characters with `string.printable`, which only works for printable ascii characters. That is good if you know you are only going to have English language text. 

Enabling unicode matching is better in theory as you get English + any other language characters, but that also means that you get a lot of garbage too (especially if you are primarily looking for English text). 

The description in the plugin says ASCII & Unicode both, but it was only getting ASCII earlier. 

The current code keeps old behavior (ASCII only), but you can toggle to full Unicode (which automatically gets you ASCII too) on line 38/39.

### DIFF
--- a/scripts/artifacts/walStrings.py
+++ b/scripts/artifacts/walStrings.py
@@ -1,56 +1,47 @@
 import os
-import textwrap
-import datetime
-import sys
 import re
 import string
+
 from pathlib import Path
 from html import escape
 
 from scripts.artifact_report import ArtifactHtmlReport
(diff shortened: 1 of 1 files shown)


abrignoni:
Will test in the next few days and commit. I apologize for the super long delay.

PR #157, opened 2020-12-22
Added a missing import

any333:


abrignoni:
I swear I had fixed it. Guess I didn't commit the changes. Thanks for catching the oversight. Idea was to limit the list of locations at the beginning of the report since the detailed locations are part of the fields in the grid.

PR #160, opened 2020-12-26
Fix bug - not displaying all pics in app snapshots

ydkhatri:
This was because certain characters in filepath need to be encoded for HTML display and certain others for making valid URLs. Fixed now.


abrignoni:
👏 👏 👏 👏 👏
Going to be reviewing some of the code for self learning purposes. Thanks!

PR #161, opened 2020-12-29
Extend mobileInstall and fixed a typo

any333:


abrignoni:
Thanks for the addition. Can't believe I missed adding the timestamp. Thank you for noticing and fixing. You rock.

PR #167, opened 2021-01-12
Added a new artifact - Keyboard Dynamic Lexicon

any333:


abrignoni:
Awesome artifact. Is there a reason to have whole file path and filename in the data portion of the report even though the same path and filename are already at the start of the report?

Screen Shot 2021-01-12 at 8 42 13 AM

any333:
I was not sure if it is better to display only the path to the folder on top and only the name of the file in the "Filename" column, or to show everything on both places. As there can be many dynamic-lexicon.dat files, depending on how many keyboard languages are used (e.g. I have en-dynamic.lm and de-dynamic.lm), it might be important to know which words are coming from which keyboard exactly.

abrignoni:
If all data files are in the same directory I would put only the directory
path at the top of the report. Then have the filenames in the data section
with the words. I think it might make it look cleaner. Thoughts?

any333:
The data files are all in the Keyboard directory, but in different folders.
Example:
.../private/var/mobile/Library/Keyboard/de-dynamic.lm/dynamic-lexicon.dat for German
and
.../private/var/mobile/Library/Keyboard/en-dynamic.lm/dynamic-lexicon.dat for Enlisch

It can be a good solution to put only the directory path at the top, as you suggest, and to put en-dynamic.lm/dynamic-lexicon.dat, de-dynamic.lm/dynamic-lexicon.dat, etc. in the data section.

abrignoni:
I like it. 👍

ydkhatri:
> As there can be many dynamic-lexicon.dat files, depending on how many keyboard languages are used (e.g. I have en-dynamic.lm and de-dynamic.lm), it might be important to know which words are coming from which keyboard exactly.

Then you can put the rest of the path at the top, and the base filename in the table. So it would look like this:

Keyboard dynamic Lexicon located at: xxx/yyy/xxd/dfdfd/dfdfdfdfdfd/dfdfdsfdfd/dfddfdf/

Filename | Found Strings
----------|---------------
en-dynamic.lm/dynamic-lexicon.dat | Hello <br>World
de-dynamic.lm/dynamic-lexicon.dat | Der

ydkhatri:
Please switch the order of columns, lets have the important stuff first.

ydkhatri:
Hey @any333 , while you are at it, you might want to parse this plist too (as a separate module). It's in the same folder - 
image

_Looks like it's dates and times of keyboard usage in apps. At least goes back a few days._

any333:
> Hey @any333 , while you are at it, you might want to parse this plist too (as a separate module). It's in the same folder -
> image
> 
> _Looks like it's dates and times of keyboard usage in apps. At least goes back a few days._

@ydkhatri I also had that thought, but was not sure about it.
I will take a look at it in the upcoming days.

any333:
@abrignoni, @ydkhatri I completed the requested changes and added a new artifact. Let me know if some modifications are needed.

abrignoni:
Merged. Very, very, very nice.

ydkhatri:
@any333 , can you look into #172

PR #170, opened 2021-01-13
SMS chat rendering issues

tobraha:
I ran into a few minor SMS chat rendering issues when working with multiple iTunes backups that were decrypted using @jfarley248's iTunes Backup Reader.  Found that I had to decrypt to 'filesystem' rather than a decrypted iTunes backup as iLEAPP does not extract SMS attachments; I intend to open a new issue (and hopefully a PR if I can wrap my head around that functionality) to address this.

### DIFF
--- a/scripts/chat_rendering.py
+++ b/scripts/chat_rendering.py
@@ -158,6 +158,7 @@ mimeTypeIcon = {
     "video":"🎥",
     "animated":"🎡",
     "application":"📎",
+    "text":"Ŧ"
 }
 
 """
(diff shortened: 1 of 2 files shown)


abrignoni:
Think you could send a PR adding your information as a contributor in the version.py file?
Thank you so much for contributing.

tobraha:
Of course - thank you _all_ for the great work you do.  I love this tool!

And it would be my genuine pleasure to be on that list; I'll submit a PR :+1:

scripts/chat_rendering.py
```diff
@@ -188,14 +189,14 @@ def integrateAtt(rec):
               <source src="{0}" type="{1}">
               <p><a href="{0}"></a> </p>
             </audio>
-            """.format(rec["content-type"])
+            """.format(rec['file-path'], rec["content-type"])
```
ydkhatri, line 192: This doesn't seem right. There are two variables that need to be filled here. This will fail on execution of this line!

ydkhatri (commented):
The changes to chat_rendering.py will cause a run time exception.

tobraha:
Had me worried there for a moment!

I did just realize that the catch in d799c28 for missing attachments isn't placed correctly (only catches on non-Windows systems; needs to be moved above 'is_platform_windows()' ).  I can send another PR if you'd like :+1:

PR #176, opened 2021-01-14
datetime module imported to bluetooth.py

bardurc:
bluetooth.py is missing datetime module.
Paired Bluetooth is therefore not parsed.
Adding "import datetime" fixes the issue.

Screenshots show output without and with datetime imported.

without datetime
with datetime


ydkhatri:
Thx for the bug fix.

PR #179, opened 2021-01-14
Fix sms attachment file check

tobraha:
Refer to #170 ; move the check for valid attachment file to include Windows.


ydkhatri:
Can you make another change along with this? Let's try to use some existing code functions like ilapfuncs.sanitize_file_name() on lines 71-72. It uses regex and is a much faster replace.
Just make the new commit, it should automatically get added to this pull.

ydkhatri:
Thanks.

tobraha:
Of course!  It looks good to me, but I didn't get a chance to test this on Windows.

PR #180, opened 2021-01-14
Added Notes artifact and optimized Reminders

any333:


abrignoni:
Very nice. Was able to test with notes. Don't have a test image with reminders.

Important tweak:
The code as is does not take into account any wal or journal files that might be part of the database.
I have a lot of artifacts that I need to modify for the same reason. One of my goals for 2021.

The new coding pattern to take them into account is as follows:

1) In the regex section get all files associated with the database.
For example instead of: 
'notes':('Notes', '*/NoteStore.sqlite'),
Do this: 
'notes': ('Notes', '*/NoteStore.sqlite*'),

2) Filter the files_found list to open only the non journaling file for querying. Be aware that as long as the journaling files are with the database python will take them into account. A sample coding pattern is as such:

Screen Shot 2021-01-15 at 11 44 43 AM

I am merging the code. If you can, add the wal / journal file piece.

Thank you so much for your contributions. This broadens the project and makes it so much better. A lot of people around the world use it daily. Again, thank you so much.

ydkhatri:
Also folks, let's not forget to update the version every time we change something. This is in scripts/version_info.py. 
Format is _major.minor.micro_
For every module addition, bump up the minor version.
For every bug fix or other minor change, bump up the micro version.

PR #182, opened 2021-01-17
Added a new artifact (Alarms) + modified Reminders and Notes to extract wal/journal/shm and parse .sqlite

any333:


abrignoni:
Getting the following exception:

Alarms [alarms] artifact executing
Reading Alarms artifact had errors!
Error was 'MTSleepAlarm'
Exception Traceback: Traceback (most recent call last): File "/Users/abrignoni/Downloads/iLEAPP-alarms/scripts/ilap_artifacts.py", line 264, in process_artifact method(files_found, report_folder, seeker) File "/Users/abrignoni/Downloads/iLEAPP-alarms/scripts/artifacts/alarms.py", line 26, in get_alarms for sleep_alarms in pl['MTAlarms']['MTSleepAlarm']: KeyError: 'MTSleepAlarm'

It seems this key is not being checked by your check_if_key_exists() function.
Thoughts?

ydkhatri:
When reading from a dictionary, it's best to retrieve key values using the 
```dict.get('key_name', val_returned_if_key_not_present)``` 
function. This won't throw any exceptions, and you can return empty strings or empty dict too.

abrignoni:
Nice!

ydkhatri:
Here is an example from ileapp repo that I modified few days back to fix a similar issue.
https://github.com/abrignoni/iLEAPP/blob/32720a9482c6428edcadc5957c7a34eb1e82385d/scripts/artifacts/appItunesmeta.py#L24

any333:
@abrignoni Thought that the keys 'MTAlarms' and 'MTSleepAlarm' are always there, so didn't implement a check initially. It should be fine now.
@ydkhatri Thanks, it's a handy solution :)

PR #193, opened 2021-02-18
Extended appleWalletTransactions

any333:


abrignoni:
Awesome! Have you tried the KML function? If you use latitude and longitude
as the names and feed it to the function in creates the mapping files
automatically.

Think you could add it? Let me get you an example from another artifact...

abrignoni:
Check out tileappdb.py artifact. Notice the functions at the end. Both timeline and kmlgen. Only other addition is the import at the top in order to use them.

When used they produce an integrated sqlite database of all geolocation points and kml files. Also a timeline sqlite database. I think it is pretty neat cause it aggregates the data across all artifacts. The databases will be in 2 folders at thr root of the report folder when processing is done. 

Hope the previous made sense. 

Screenshot_20210218-175653

abrignoni:
Ignore previous messages. I can't github today. 😂

abrignoni:
Just checkout the kmlgen function. Name the geolocation fields latitude and longitude for it to work. 

Screenshot_20210218-175653

PR #221, opened 2021-04-05
Repaired a broken query for Health Distance

any333:


abrignoni:
You are the best. Thank you!!

PR #247, opened 2021-09-25
Add Venmo Parser

jfarley248:
### Venmo Parsing
Venmo artifact added, gets transaction history PrivateFeed, PublicFeed, and FriendsFeed, also enumerates user data from them. Outputting two new reports, one for transactions and one for users

usuwn9KTmq

^ Example of transaction history report

chrome_00JAflTJXw

^ Example of  user report

### Some things to note:

This does not grab comments added to payments, nor parses 'disbursements' from Venmo, for some reason venmo randomly pays people sometimes. This can be easily added.

Unsure how to add a custom little icon for the html report

Unsure if this should be split into two 'artifacts' one for users and one for transactions. But can be easily fixed if so

For the filename located I used the list of the 3 files used, although each row doesn't specify which file it came from exactly. Wasn't too sure how to specify different locations for similar data - or it can be split into 3 diff reports like public, private, and friends transactions to fix this as well.

Also unsure if this follows the coding practices for this project completely, but let me know if something should be fixed/cleaned up :)


stark4n6:
In "scripts/reports.py" you can add a line for changing the icon. You can find a list of icons here https://feathericons.com/. Something like "dollar-sign" would probably work.

Issue #257, opened 2021-11-12
Consider numpy version bump


jijames:
iLEAPP requirements.txt set numpy == 1.19.3. This numpy version does not appear compatible with python3.10.

Installed numpy-1.21.4 and iLEAPP appears to be working.

threeplanetssoftware:
This issue should be able to be closed, based on the merged PR listed above.

PR #267, opened 2022-01-28
Updated requirements.txt to bump numpy versionand fix python-magic

### LINKED ISSUE #266: Switch to python-magic-bin breaks Linux support
Commit 10197b6d34271842c2e21b0f23cbc8009d6bf1dc updated `requirements.txt` to use python-magic-bin, instead of python-magic. There is [no Linux whl file](https://pypi.org/project/python-magic-bin/#files) for python-magic-bin, so Linux users will run into this error the next time they attempt to install requirements:

```
ERROR: Could not find a version that satisfies the requirement python-magic-bin==0.4.14 (from versions: none)
ERROR: No matching distribution found for python-magic-bin==0.4.14
```

This should be solvable by checking the platform and continuing to include python-magic if the platform is not Windows or Mac. I'll submit a PR.

threeplanetssoftware:
Fixes #266 and #257 by bumping the numpy version to be compatible with python 3.10 and ensuring that Linux users still fall back on python-magic, not the python-magic-bin package. I haven't been able to extensively test this, it would be worthwhile to make sure it works appropriately on Mac since I think you use them more than me. I've just tried it on Linux and it fixed the issues immediately.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-numpy==1.19.3
+numpy==1.21.4
 astc_decomp
 biplist
 bs4
@@ -15,4 +15,6 @@ biplist
 simplekml
 pandas
 pycryptodome
-python-magic-bin==0.4.14
+python-magic==0.4.24; platform_system == "Linux"
+python-magic-bin==0.4.14; platform_system == "Windows"
+python-magic-bin==0.4.14; platform_system == "Darwin"


abrignoni:
I didn't know one could constrain requirements by OS. Very cool. Learned something new today. Thanks!

abrignoni:
Was this tested on Windows? The extreme majority of the user base is Windows based.

threeplanetssoftware:
I haven't yet tested on Windows, I've been fighting to get my server back operational so none of my test VMs are running yet. Trying a new power supply was literally the next thing on my to-do list tonight. I just tried it on my Linux dev box which has python 3.10.

threeplanetssoftware:
Confirmed that for Linux this pulls python-magic and for Windows it pulls python-magic-bin. Running a report on one of my test extractions appears to work similarly on both. Haven't tested Mac yet, but it should behave like Windows.

abrignoni:
Pure awesomeness!

PR #290, opened 2022-04-16
Add Bumble Parser (Preliminary)

stark4n6:
- Chat messages
- Local account details


stark4n6:
Need to add header and bump version, closing for now

PR #313, opened 2022-06-14
Update geodMapTiles.py

Krypterry:
Minor update to include check/support for older iOS DB schema.


abrignoni:
Thank you, thank you, thank you!

PR #438, opened 2023-03-26
Update Health.py

SQLMcGee:
This change fixes the Celsius temp column which pulled from the Average Heart Rate (BPM) column. Also added column within Health - Workouts for Total Time Duration. Total Time Duration reviewed side by side with Workout Duration can show variations which could be significant within an investigation/examination Additional details on total time versus workout time with published work "Enriching Investigations with Apple Watch Data Through the healthdb_secure.sqlite Database" at https://dfir.pubpub.org/pub/xqvcn3hj/release/1

### DIFF
--- a/scripts/artifacts/Health.py
+++ b/scripts/artifacts/Health.py
@@ -13,7 +13,9 @@
 # Date: 2023-03-02
 
 # Updates: @SQLMcGee
-# Date: 2023-03-21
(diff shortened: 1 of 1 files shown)


abrignoni:
Thank you so much for all your work on these. The user community will make good use of it. Me being the first one. Thank you. 🙏

PR #470, opened 2023-05-16
Update ileapp.spec

markmckinnon:
Add chats.css to datas.
Fix formatting as well

Version 1.18.6 command line program will need to be compiled as it does not work


abrignoni:
Will make a new binary tonight.

PR #501, opened 2023-08-09
Update vippsContacts.py

thisislola:
Update. ZPHONENUMBERS has been changed to ZRAWPHONENUMBERS in the Vipps database.


abrignoni:
Could does_column_exist_in_db() from ilapfuncs.py be used to verify which table is present in order to use the appropiate query? The idea is to ensure backward compatibility with databases that use the old column name.

PR #503, opened 2023-08-09
Update vipps.py

thisislola:
Update on Vipps Transaction. The BLOB/JSON is different if it's a message or a payment. Payments are also showing the decimals/cents, however Vipps doesn't accept/send payments with cents (øre).


abrignoni:
Could the code not include a class and decorators? The idea is to make the code easily accessible to beginner to medium experienced digital forensics examiner.

PR #541, opened 2023-10-09
Implement and Adopt New Self Documenting Artifact Info Structure v2

JamesHabben:
Some of the modules have comments at the top of the code to share information about what the module does, who wrote it, release date, and more. I've created a new format of the artifact structure that was at the bottom of the code, and included the ability for the comments and the artifact info to be combined into one structure and placed at the top of the code in a self documenting type of fashion. I then updated the module loaded to use v2 but kept backwards compatibility with modules not updated to v2.

- Introduced a new artifact structure `__artifacts_v2__` to replace the previous `__artifacts__` structure, enhancing readability and organization of metadata.
- Updated the plugin loader to accommodate the new `__artifacts_v2__` structure while retaining backward compatibility with the old `__artifacts__` structure.
- Converted a few existing modules to the new `__artifacts_v2__` format to demonstrate the improved structure and serve as examples for future conversions.
- Updated the README.md to reflect the new `__artifacts_v2__` structure and provided guidelines for contributing plugins using the new format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,27 +84,49 @@ $ python ileapp.py --help
 
 ## Contributing artifact plugins
 
-Each plugin is a Python source file which should be added to the `scripts/artifacts` folder which will be loaded 
-dynamically each time ILEAPP is run.
+Each plugin is a Python source file which should be added to the `scripts/artifacts` folder which will be loaded dynamically each time ILEAPP is run.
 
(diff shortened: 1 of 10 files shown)


JamesHabben:
Sorry, I am not very good with GitHub stuff and renamed my branch after creating this.

PR #549, opened 2023-10-11
ATXDataStore

Magpol:
Parses ATXDataStore and matches actions with Frequent locations, when available.


abrignoni:
Well done! Thank you so much!!!!
Added icon to report, KML map for the geolocation points, renamed and reorder a few columns.

PR #572, opened 2023-10-18
refactor report categories

JamesHabben:
refactored the icons into a dictionary object and kept the search ability by including a "_mode" key. first pass is a direct lookup for an exact match of the key. if that isn't found, it moves to search mode with a reduced set of records based on those with a `_mode` : `search` in their record. if search mode is declared, both category and artifact keys are searched. goal here would be to make this file the same across all xLEAPP versions. if someone contributes a new icon mapping in one xLEAPP, then it would be simple to do a diff merge into all the other xLEAPP versions.

function `def sort_and_print_mappings():` reads in the `icon_mappings` object and prints it back out sorted to console

its called by the `if __name__ == '__main__':` if you run the `report.py` file as a standalone script.

Then you can copy paste back into the script from the console and the object will be sorted.

### DIFF
--- a/scripts/report.py
+++ b/scripts/report.py
@@ -10,456 +10,438 @@ from scripts.html_parts import *
 from scripts.ilapfuncs import logfunc
 from scripts.version_info import aleapp_version, aleapp_contributors
 
+# Icon Mappings Dictionary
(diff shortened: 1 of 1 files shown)


abrignoni:
This looks so clean and fantastic. Thank you so much!

PR #643, opened 2023-12-22
Fix relative path for pyinstaller

stolenfootball:
Pyinstaller currently fails with ileapp.spec because of the lack of a dot in front of one of the paths in the datas array.


abrignoni:
Thank you!

PR #669, opened 2024-02-08
Waze

djangofaiola:
account, session info, searched locations, recent locations, favorite locations, share locations, text-to-speech navigation and track GPS quality.

### DIFF
--- a/scripts/artifacts/waze.py
+++ b/scripts/artifacts/waze.py
@@ -1,0 +1,607 @@
+__artifacts_v2__ = {
+    "waze": {
+        "name": "Waze",
+        "description": "Get account, session, searched locations, recent locations, favorite locations, "
+					   "share locations, text-to-speech navigation and track GPS quality.",
+        "author": "Django Faiola (djangofaiola.blogspot.com @DjangoFaiola)",
+        "version": "0.1",
+        "date": "2024-02-02",
+        "requirements": "none",
+        "category": "Waze",
+        "notes": "",
+        "paths": ('*/mobile/Containers/Data/Application/*/Documents/user.db*',
+                  '*/mobile/Containers/Data/Application/*/Documents/user',
+                  '*/mobile/Containers/Data/Application/*/Documents/session',
+                  '*/mobile/Containers/Data/Application/*/Documents/spdlog.*logdata',
+                  '*/mobile/Containers/Data/Application/*/Library/Caches/tts/tts.db*'),
+        "function": "get_waze"
+    }
+}
+
+import os
+import re
+import pathlib
+import shutil
+import sqlite3
+import textwrap
+import datetime
+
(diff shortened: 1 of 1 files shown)


abrignoni:
Thank you for the addition. Just one detail, the regex for the folders might be a little broad. One way I limit is by using  .com.apple.mobile_container_manager.metadata.plist to limit the extracted files and folders to the app directory in question. I have an example her: https://github.com/abrignoni/iLEAPP/pull/652/files

Thoughts?

djangofaiola:
Ciao, do you want something like this?
```
        "paths": ('*/mobile/Containers/Data/Application/*/Documents/user.db*',
                  '*/mobile/Containers/Data/Application/*/.com.apple.mobile_container_manager.metadata.plist'),
#                 '*/mobile/Containers/Data/Application/*/Documents/user',
#                 '*/mobile/Containers/Data/Application/*/Documents/session',
#                 '*/mobile/Containers/Data/Application/*/Documents/spdlog.*logdata',
#                 '*/mobile/Containers/Data/Application/*/Library/Caches/tts/tts.db*'),


# waze
def get_waze(files_found, report_folder, seeker, wrap_text, timezone_offset):
    for file_found in files_found:
        # .com.apple.mobile_container_manager.metadata.plist
        if file_found.endswith('.com.apple.mobile_container_manager.metadata.plist'):
            with open(file_found, 'rb') as f:
                pl = plistlib.load(f)
                if pl['MCMMetadataIdentifier'] == 'com.waze.iphone':
                    fulldir = (os.path.dirname(file_found))
                    identifier = (os.path.basename(fulldir))

                    # user
                    path_list = seeker.search(f'*/{identifier}/Documents/user', True)
                    if len(path_list) > 0:
                        get_account(path_list[0], report_folder, timezone_offset)

                    # session
                    path_list = seeker.search(f'*/{identifier}/Documents/session', True)
                    if len(path_list) > 0:
                        get_session(path_list[0], report_folder, timezone_offset)

                    # tts.db
                    path_list = seeker.search(f'*/{identifier}/Library/Caches/tts/tts.db', True)
                    if len(path_list) > 0:
                        get_tts(path_list[0], report_folder, timezone_offset)

                    # spdlog.*logdata
                    path_list = seeker.search(f'*/{identifier}/Documents/spdlog.*logdata')
                    if len(path_list) > 0:
                        get_gps_quality(path_list, report_folder, timezone_offset)

                    break

        # user.db
        elif file_found.endswith('user.db'):

```

abrignoni:
Yes. My concern is that user and session could be so common there you could get a lot of false positive files pulled in unless  you narrow it down to the directory where Waze is.

What do you think?

djangofaiola:
I apologize for my English, it is very bad.
You are right, I didn't think of this problem.
I am not a Python programmer and I hope I used the functions of iLEAPP correctly.
thanks for the help

PR #705, opened 2024-03-29
Inconsistent join on Z_PK in DataUsage.sqlite

Schramp:
netusage incorrectly joined two tables of DataUsage.sqlite on Z_PK. Compared to code in
./scripts/artifacts/DataUsage.py

### DIFF
--- a/scripts/artifacts/netusage.py
+++ b/scripts/artifacts/netusage.py
@@ -32,7 +32,7 @@ def get_netusage(files_found, report_folder, seeker, wrap_text, timezone_offset)
             ZLIVEUSAGE.ZWIREDIN,
             ZLIVEUSAGE.ZWIREDOUT
             from ZLIVEUSAGE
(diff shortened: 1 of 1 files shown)


Schramp:
Sorry for the late update, while inspecting the diff I noticed a vim artefact that added a trailing newline

stark4n6:
@Schramp I tried your updated query on a test CTF image and it failed, are you sure your query is correct?

stark4n6:
@Schramp do you mean at line 136 instead?

stark4n6:
@Schramp I'll get this patched up, I think I made both parsers and combined both in the NetUsage at some point, not realizing the other was already here. So what I'll do is remove the DataUsage section from the NetUsage.py and it'll be two separate parsers now.

I had found some small other bugs in the NetUsage script as is so I'll push all of it over shortly to the main repo. Going to close this PR out.

PR #731, opened 2024-05-06
Updates to parsed embedded files plists and jpgs 

ScottKjr3347:
updates to photos.sqlite parsers for embedded plists and embedded preview jpgs.


abrignoni:
I agree. Think you could make the exclusion? If not I will look at the code and try to find where it is done now after the GUI changes. 🙂👍

Johann-PLW:
Sure.
I know the code a lot better now 🤪 and know exactly where it is done.
I'm also going to check all the modules.

ScottKjr3347:
Thanks for looking at these ph94-97 should be excluded by default. I just haven't found the file to change the included/exclude them by default yet.

Johann-PLW:
Thank you very much @ScottKjr3347 
I am allso going to exclude by default Ph94-97.
The modules to exclude by default are indicated in ileapp.py and ileappGUI.py scripts. But I like the idea to create a separate file to do that. I will see what I can do

PR #740, opened 2024-05-16
Added support for mbdb manifests

Mori19:
Old iTunes backups dont seem to work at the moment as they use the Manifest.mbdb instead of the Manifest.db. Support has been added through a new seeker type to allow parsing old backups.

Mostly from this discussion here: https://stackoverflow.com/questions/3085153/how-to-parse-the-manifest-mbdb-file-in-an-ios-4-0-itunes-backup
Changed to fit iLEAPP

### DIFF
--- a/ileapp.py
+++ b/ileapp.py
@@ -139,7 +139,7 @@ def main():
                               "'fs' for a folder containing extracted files with normal paths and names, "
                               "'tar', 'zip', or 'gz' for compressed packages containing files with normal names, "
                               "or 'itunes' for a folder containing a raw iTunes backup with hashed paths and names."
-                             "If you have an older iTunes backup using the Manifest.mbdb format instead, us 'itunes-mbdb'.))
+                             "If you have an older iTunes backup using the Manifest.mbdb format instead, us 'itunes-mbdb'."))
     parser.add_argument('-o', '--output_path', required=False, action="store",
                         help='Path to base output folder (this must exist)')
     parser.add_argument('-i', '--input_path', required=False, action="store", help='Path to input file/folder')
--- a/ileapp.py
+++ b/ileapp.py
@@ -134,11 +134,12 @@ def create_casedata(path):
 
 def main():
     parser = argparse.ArgumentParser(description='iLEAPP: iOS Logs, Events, And Plists Parser.')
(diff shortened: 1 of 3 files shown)


Johann-PLW:
Hello @Mori19,

Thank you very much for adding the support for iTunes backups from older iOS versions.

I have tested with many different backups (iOS 7.x, iOS 8.x). This works for some of them, but for a large number, I get the following error:
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xba in position 10: invalid start byte`

The byte and the position both vary depending the manifest.mdbd parsed.

Let me know if you need one of these stubborn manifest.mdbd files to investigate.

Regards.

Mori19:
Hi @Johann-PLW, 

I would appreciate some of the examples of non working mbdb files. 

Hopefully I should be able to get those sorted out once I've had a look through them. 

Cheers

Johann-PLW:
Hi @Mori19 

Sorry for the late reply but you will find in the following zip a working and a non-working Manifest.mbdb of the same iPhone;
Let me know if you need more examples after this one;

Greetz

[Manifests.mbdb.zip](https://github.com/user-attachments/files/15744120/Manifests.mbdb.zip)

Mori19:
Hi @Johann-PLW, 

No worries on that at all. 

I think I found the problem - some of the files have additional binary properties listed in the manifest.mbdb. As the only actual data used is the filename and domain, I have updated the parser to not attempt to decode the property, and instead ignore it. 

This resulted in no change to the functionality, but resolved the error message when I was looking at it. 

Please let me know if the issue still persists. 

Cheers

Johann-PLW:
Hi @Mori19,

Thank you very much for your update.
I've tested 12 old iTunes backups and the issue was gone except for this Manifest.mbdb :
[Manifest.mbdb.zip](https://github.com/user-attachments/files/15859684/Manifest.mbdb.zip)

Cheers

Mori19:
Hi @Johann-PLW,

Looks like it was the same issue as before, but in the encryption key field instead of the properties. 

I've stopped all decoding of non-essential fields, so this should be ok now. 

Cheers

Johann-PLW:
Everything works fine now.
Thank you very much for your contribution.

PR #744, opened 2024-05-20
Create Watch Worn Data.py

MetadataForensics:
This artifact provides an "at a glance" review of time periods in which the Apple Watch is worn. This data can lend to pattern of life analysis as well as providing structure to periods in which data such as heart rate data will be generated and recorded.


Johann-PLW:
Thank you very much @MetadataForensics for adding this new Health artifact in iLEAPP.

I have just:
- removed the spaces in the filename to be consistent with the other parsers and to be more complient with Style Guide for Python Code (pep 8),
- removed unused imports
- and used your comments to migrate to artifact v2

PR #748, opened 2024-05-30
Google Translate

djangofaiola:
History, Favorite translations and Text-To-Speech

### DIFF
--- a/scripts/artifacts/googleTranslate.py
+++ b/scripts/artifacts/googleTranslate.py
@@ -1,0 +1,202 @@
+__artifacts_v2__ = {
+    "googleTranslate": {
+        "name": "Google Translate",
+        "description": "History, Favorite translations and Text-To-Speech",
+        "author": "Django Faiola (djangofaiola.blogspot.com)",
+        "version": "0.1.0",
+        "date": "30/05/2024",
+        "requirements": "none",
+        "category": "Google Translate",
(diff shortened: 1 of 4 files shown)


Johann-PLW:
Hello @djangofaiola,

Thank you very much for these additions.
I got this exception with Google Translate module on my own iPhone (iOS 17.4.1) with Google Translate 8.9.0:
`Exception while parsing Google Translate Text-To-Speech: no such table: tts`

Which versions of Google Translator have you tested?

djangofaiola:
>`Exception while parsing Google Translate Text-To-Speech: no such table: tts`
Hi,
In older versions 5.x, 6.x the tts table is present but in newer 8.x it is absent.
I fixed the code with a check if the tts table exists.
 
> Which versions of Google Translator have you tested?
5.18.59063, 6.4.59163, 8.3.0

Johann-PLW:
Hi,
As mentionned in my previous message, it is Google Translate v8.9.0, the current one available on the App Store.
Even if I used it, my translate.db database is desperately empty.
I have some older backups and I will have a look.

Thank you again.

Johann-PLW:
Hi again @djangofaiola,

FYI, the tts table was still present in version 8.1.1
Unfortunately, I don't have any intermediate version to find out from which version this table was removed.

PR merged.

PR #797, opened 2024-08-01
Create WatchSleepData.py

MetadataForensics:
These artifacts provide an 'at a glance' review of sleep periods when the Apple Watch is worn, given required user settings. This data is broken into Sleep States, following Apple's use of them, and breaks down the data to mirror the Health application review of Sleep. Additional details published within 'Sleepless in Cupertino: A Forensic Dive into Apple Watch Sleep Tracking'.


MetadataForensics:
Health Sleep Preview 0
Health Sleep Preview 1

PR #803, opened 2024-08-21
Update messageRetention.py

stark4n6:


JamesHabben:
@stark4n6 what do you think about adding a path for itunes backups?
image

JamesHabben:
well, after I looked again I realized you actually did just that. sorry!

stark4n6:
@JamesHabben it should encapsulate it, but I might have missed the other file version (lowercase mobileSMS)

PR #819, opened 2024-09-24
Added support for the App Withings Health Mate

kalink0:
- Added parser for the App Withings Health Mate
- Decoding based on Post on my blog: https://bebinary4n6.blogspot.com/2024/09/withings-healthmate-on-ios.html


stark4n6:
@kalink0 can you check if this regex path is too broad?

image

kalink0:
> @kalink0 can you check if this regex path is too broad?
> 
> image

Hi, I know what you mean - could be problem.
Well, I cannot narrow it down any further - I now added the base path for sandbox pathes (/private/var/mobile/Container/Data/Application/) in front of the regex. So the file only matches when it is in the Sandbox Path on iOS.
But additionally: I always test the regex (or more specific the whole script) on different iOS FFS zips (my own and the one from Binary Hick) and didn't got any wrong matches. Hope that's fine.

stark4n6:
@kalink0 was just wondering if it would get any false positives based off the file name, but should be okay.

One thing of note, if you have any artifacts that have a timestamp and you want to use the timelining report feature, the timestamp column should be listed first in the append to the data_list, can you fix that up quick along with the report headers to reflect that as well? Other than that looks good.

kalink0:
@stark4n6 sure, i've done that quickly.
thx for your feedback. Good to know with the timeline - will keep that in mind for other artifacts ;-)

stark4n6:
@kalink0 merged, thanks for your contribution!

PR #943, opened 2024-11-20
Fix bug in telegramMessages.py

kobo220:
Current artifact plugin is missing several Types in the TelegramMediaAction class. Currently, if the plugin encounters any of these missing Types an error is thrown and no report is generated. The missing types were added. Additionally error handling was added to account for any possible missed or future added Types. These types will default to type 'unknown' in the report and will output an error message to terminal.


kobo220:
Incorrectly submitted PR. Will try again.

PR #975, opened 2024-12-13
Trusted Peers Database

charpy4n6:
Parses Information about Devices Associated with iCloud Account. Information includes timestamp, Model, Model Version, Device Name, Serial Number and Passcode Length of the device. Location /private/var/Keychains/com.apple.security.keychain-defaultContext.TrustedPeersHelper.db


Johann-PLW:
Hi @charpy4n6,
Your module can be merged as it is but I have a question.
There are a lot of duplicates in the report. This is the way the information is recorded in the database but there are only 24 unique records out of 1746 (checked with TSV export on Excel and Numbers) with Josh's iOS 17 image.
If you think:
- we have to report all the records, I can merge your PR as it is;
- we can only display the unique ones, you can simply add DISTINCT to your query before the datetime function.

charpy4n6:
Agree. My test data actually only had 2 unique. I debated if we should show
all that are in DB or not. Whatever you think is good. I can fix it to only
show unique.
Heather Charpentier
NYSP
***@***.***

Johann-PLW:
In my opinion, I think it's better for readability to only show unique reports.
Johann

charpy4n6:
Ill do it today :)

charpy4n6:
Done! Tested it too and it works! Thank you!!!

PR #1001, opened 2025-01-04
Refactor ileapp.spec for improved configuration and clarity

shamrockmuffin:
- Updated the Analysis configuration to remove hardcoded paths and unnecessary parameters.
- Simplified the PYZ and EXE definitions for better readability and maintainability.
- Removed unused options and added placeholders for future configurations.


JamesHabben:
I notice that `simplekml` is added to hidden but it wasn't there before. Can you share more about what this update is doing and why?

PR #1023, opened 2025-01-19
Add native support for iTunes encrypted backups

threeplanetssoftware:
Good evening!

This PR is an initial shot at adding support for encrypted iTunes backups directly in iLEAPP, instead of needing to use an external tool to unpack it first. Python is certainly not my preferred daily driver language, so I'm more than open to hearing improvements you'd like to see before merging. That said, I'm also plenty happy for someone else to run with this, it was just a quick Saturday night feature push to clear my head of another technical hurdle I was stuck on. 

This includes new command line line switches to either receive the password (`--itunes_passcode`) or prompt for it (`--prompt_for_passcode`). It changes the iTunes FileSeeker to automatically identify if the file is encrypted by reading `Manifest.plist`, handle decrypting the Manifest.db, building the set of protection classes that will unwrap encryption keys, and decrypting files as they get copied out. These changes try to be as nice as possible to RAM and CPU by only grabbing additional details if the folder is actually encrypted, for example. It also sets a default of `None` to the function arguments it added, so no current calls should need to be updated.

This does NOT yet touch the GUI which still needs a way to prompt for the password. Because FileSeeker detects the encrypted backups automatically, once there is a way to put a passcode into the GUI, it should be minimal changes to make that work, as well. 

Having run a comparison between an encrypted and unencrypted backup from the same device (unfortunately, not from the same time, so there are some differences between them) showed largely similar results. The encrypted backup too 5.45 seconds of processing time and the unencrypted took 3.12, so for large devices total time might chew up a bit with all the decryption, but it is hard to say exactly how much since the extra 2 seconds includes a million-round PBKDF2 iteration and it may be once that overhead is done the files are nearly instantaneous.

Appreciate any feedback you have. Would be happy to work on this as I have spare time, or see someone make edits to it directly to get it merged.


stark4n6:
@threeplanetssoftware I know @Johann-PLW had put some more code for iTunes parsing so you might have to review the current base code and change as needed, much appreciated!

stark4n6:
@threeplanetssoftware no need to apologize, thanks for hopping right back into this PR to fix it months later. I'll try to do more testing, and will work with @Johann-PLW to get something for passcodes in GUI then we can merge I think.

PR #1031, opened 2025-01-23
Update nsVault.py for lava output

Johann-PLW:


Johann-PLW:
Hi @charpy4n6 
Could you check if this update works for lava output?
I do not have test data.

Many thanks

PR #1041, opened 2025-01-30
Add files via upload

MetadataForensics:
This provides an in-depth review of all Apple devices and their OS version timeframes parsed from the Health Application. Adds parsing support for Device and OS Timeframe, Device Isolation (both iOS 15, 16, 17, and 18), and Current Model and OS (iOS 18 only) for Apple devices, pulling data from the healthdb_secure and healthdb databases.  Additional information is available here: https://metadataperspective.com/2025/01/30/beyond-the-logs-using-the-health-app-to-uncover-device-model-and-os-history/


stark4n6:
Hi James (@MetadataForensics), you may be able to utilize the Build ID list we use to determine device model and iOS versioning for your queries to make it simpler, take a look [https://github.com/abrignoni/iLEAPP/blob/main/scripts/builds_ids.py](https://github.com/abrignoni/iLEAPP/blob/main/scripts/builds_ids.py)

MetadataForensics:
Right, I see how that'd work. Making that change now. I'll let you know when it's ready for another look. Thanks!

MetadataForensics:
Kevin, you know - the more I thought about it the more I like the builds_ids.. This now incorporates both the device_id and the OS_build and takes away the complex WHERE clause. Testing this against my data (and the updated Group By in this latest commit) this should now be good to go. This makes it much more simple. I started all this in DB Browser so the resources of iLEAPP are much more helpful.

stark4n6:
@MetadataForensics sorry for the delay, will try to test this afternoon. Does this have extra info parsed from the other Provenance parser in Health?

MetadataForensics:
@stark4n6 Not an issue at all, not all my comments had your handle so notifications for changes and commits were few. 

So, this does have extra info aside from the Provenance parser in Health. Information pulled within the Provenance parser is there but this utilizes timestamps to provide a full timeline of device usage and the grouping shows the OS upgrade progression in a clean way. This has three artifacts in one: full breakdown of all devices, their OS, and timeframes; isolated list of just the devices and the timeframes of use; then current devices and OS. Having this separate from the health.py is a plus too because, while this information is from the Health Application, if the iLEAPP user is seeking device information they may isolate those artifacts instead of seeking it from Health. 

Also, just notices healthWorkouts, healthProvenances, healthHeadphoneAudioLevels, healthHeartRate, and healthRestingHeartRate all fail at the same point in the respective artifacts. Seems the same fix will likely fix all of them.

stark4n6:
@MetadataForensics looks good, merged, thanks for you patience and contributions.

PR #1043, opened 2025-02-02
Fix error when attaching a SQLite DB in R/O on Windows

Johann-PLW:


stark4n6:
@Johann-PLW worked fine on my test iOS 17 image that was failing earlier

PR #1047, opened 2025-02-05
Update health.py

MetadataForensics:
Prior script was failing when attempting to select query based on iOS version. This searches the database for whether the iOS 16, 17, 18 workout_activities table is present else, runs the iOS 15 and older query. Note, otherwise the workout queries themselves were unchanged.

### DIFF
--- a/scripts/artifacts/health.py
+++ b/scripts/artifacts/health.py
@@ -12,7 +12,8 @@ __artifacts_v2__ = {
             investigation/examination. Additional details published within 'Enriching Investigations with Apple Watch Data Through the \
             healthdb_secure.sqlite Database' at https://dfir.pubpub.org/pub/xqvcn3hj/release/1",
         "paths": ('*Health/healthdb_secure.sqlite*', '*Health/healthdb.sqlite*'),
-        "output_types": ["html", "tsv", "lava"]
+        "output_types": ["html", "tsv", "lava"],
+        "artifact_icon": 'activity'
     },
     "healthProvenances": {
         "name": "Health - Provenances",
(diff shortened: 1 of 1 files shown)


stark4n6:
@MetadataForensics check [ilapsfuncs](https://github.com/abrignoni/iLEAPP/blob/main/scripts/ilapfuncs.py), there are functions that can be used in place of your changes I think:

does_column_exist_in_db

does_table_exist_in_db

MetadataForensics:
@stark4n6 Thanks for that resource, I think I have found out more about the internal resources of iLEAPP this week than ever before. I see the usage of the does_table_exist_in_db and how that could likely work for this script, it is the equivalent of what my added lines do here, but mine also flip-flops the order of the queries as the previous latter query was iOS 16 and newer and the added table is also iOS 16 and newer. So where originally the script ran off "if version.parse(iOS_version) < version.parse("16"):" then run the query for iOS 15 and older else the query for iOS 16 and newer, now it searches for the iOS 16 and newer table, runs the iOS 16 and newer query, else runs the old query. I may be mistaken, and please correct me if I am, but I believe this solution is equivalent to that resource with less lines. But you may be able to work a more elegant solution too. Thanks again.

stark4n6:
@MetadataForensics will merge for now, as it looks ok, but will test further when time permits.

MetadataForensics:
Fantastic, and thanks. Let me know if you find any issues I can assist with. Just pulled it again and no issues with my iOS 15 and iOS 18 datasets.









[cid:d26baf40-cc63-415e-a969-51662fdae333]

James R. McGee

Director of Digital Forensics Research

Senior Digital Forensic Examiner


Metadata Forensics, LLC.



Phone: 866-382-3282

Email: ***@***.***



233 12th Street, Suite 509

Columbus, GA 31901



www.metadataforensics.com<http://www.metadataforensics.com>

________________________________

PR #1048, opened 2025-02-05
Update report_icons.py

MetadataForensics:
Aesthetics only, adds icons for Height, Weight, All Watch Sleep Data, and Watch By Sleep Period.


stark4n6:
@MetadataForensics these no longer need to be added to **report_icons**, they can be added directly to the header of the parser, just add like here in the picture.

image

MetadataForensics:
Right, initially I attempted to add artifact icons for artifacts that did not have icons (neither through the health.py or the report_icons.py) straight into the health.py but it failed as not all artifact headers had an icon line. I will close this request and add to the health.py, thanks!

PR #1056, opened 2025-02-07
Update Call History Group Call 

stark4n6:
Parser was missing .py extension


stark4n6:
@MetadataForensics I ran this on Josh Hickman's iOS 17 image and it appears all the same info is being pulled into the Call History too, I presume the queries are very similar (yours had an extra column). Any way they can be merged into a single artifact? I can take a look this afternoon to see if it's feasible

MetadataForensics:
@stark4n6 I think that could be merged into a single artifact - I just didn't have a dataset old enough to have the "call" table within the old_query. Part of how I formed this one was so it mirrored the Call History artifact for merge potential with three key differences: 1) I use a group_concat aggregate function on zhandle.zvalue instead of zaddress for the phone number, with my joins to pull that column 2) the added column "Call Description" which you'd mentioned 3) iOS 18 uses the Z_4REMOTEPARTICIPANTHANDLES column whereas iOS 16 uses the Z_3REMOTEPARTICIPANTHANDLES column within the Z_2REMOTEPARTICIPANTHANDLES table. This 3rd factor is what makes it probably most complicated for merging.. I would have attempted the merge to update the original artifact instead of creating another but didn't have the data and didn't want to break the artifact for the older data I couldn't test.

MetadataForensics:
@stark4n6 Also, if merged, the interactionC of it all could be added most likely too - with probably no issues.. But again I don't have the "call" table older data. Thanks for your review and continued work on this!

stark4n6:
@MetadataForensics let me see if I can dig one out of the archive of CTF images

MetadataForensics:
@stark4n6 Not sure how far back you've been able to find.. the oldest version I have is iOS 11.4 and even then it's still the CallHistory.storedata with a ZCALLRECORD table. Granted, the current Call History artifact parses successfully and the Call History Group Call does not as there wasn't a ZHANDLE table back then, let alone the Z_2REMOTEPARTICIPANTHANDLES

PR #1083, opened 2025-03-09
Add Multiple Source Files as a List

BrunoFischerGermany:
When more then one Sources (may 2 Databasefiles) in The Report, they can be add in al List.
grafik

### DIFF
--- a/scripts/artifact_report.py
+++ b/scripts/artifact_report.py
@@ -84,12 +84,16 @@ class ArtifactHtmlReport:
         num_entries = len(data_list)
         if write_total:
             self.write_minor_header(f'Total number of entries: {num_entries}', 'h6')
-        if write_location:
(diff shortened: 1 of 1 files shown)


BrunoFischerGermany:
grafik
If the Local user is the Receiver, the Field Forwarded from is empty. Maybe there could insert the "Sender Name" and "From ID".

BrunoFischerGermany:
grafik
So is it shown.

JamesHabben:
the source file modification is straight forward enough, but there are conflicts with the current version of whatsapp artifact module. can you remove the whatsapp and we can merge the HTML edit? 

@Johann-PLW made some changes to whatsapp recently with a move to a new media manager system we are working on. i havent dug in enough to see if that is the conflict or not.

PR #1118, opened 2025-04-11
handle times before unix epoch

snoop168:
Need to manually subtract epoch time to handle pre-epoch dates.


JamesHabben:
seems similar to the issue in #1174 so maybe the code should use a common function to resolve the before time.

PR #1120, opened 2025-04-11
voicemail transcript files

snoop168:
Parse voicemail transcript files, found in BFU extractions


Johann-PLW:
Added in the voicemail module

PR #1125, opened 2025-04-14
Create storeUser parser

stark4n6:
- Parses storeUser.db for installed app history
- Parses storeUser.db for App Store purchased app history


Johann-PLW (approved):
The module works fine.
You can remove unused imports.

stark4n6:
@Johann-PLW removed the one I missed, should be good now

PR #1139, opened 2025-05-03
Update BeReal.py

djangofaiola:
Added support for Hackemist SDWebImage Cache, EntitiesStore.sqlite, Cache.db
Fixed minor bugs

### DIFF
--- a/scripts/artifacts/BeReal.py
+++ b/scripts/artifacts/BeReal.py
@@ -1,70 +1,76 @@
 __artifacts_v2__ = {
     "bereal_preferences": {
-        "name": "BeReal Preferences",
+        "name": "Preferences",
         "description": "Parses and extract BeReal Preferences",
         "author": "@djangofaiola",
-        "version": "0.1",
+        "version": "0.2",
         "creation_date": "2024-12-20",
-        "last_update_date": "2024-03-09",
+        "last_update_date": "2025-05-03",
         "requirements": "none",
         "category": "BeReal",
         "notes": "https://djangofaiola.blogspot.com",
-        "paths": ('*/mobile/Containers/Shared/AppGroup/*/Library/Preferences/group.BeReal.plist'),
+        "paths": ('*/mobile/Containers/Shared/AppGroup/*/Library/Preferences/group.BeReal.plist',
+                  '*/mobile/Containers/Data/Application/*/Library/Preferences/AlexisBarreyat.BeReal.plist'),
         "output_types": [ "none" ],
         "artifact_icon": "settings"
     },
     "bereal_accounts": {
-        "name": "BeReal Accounts",
+        "name": "Accounts",
         "description": "Parses and extract BeReal Accounts",
         "author": "@djangofaiola",
-        "version": "0.1",
+        "version": "0.2",
         "creation_date": "2024-12-20",
-        "last_update_date": "2024-03-08",
+        "last_update_date": "2025-05-03",
         "requirements": "none",
         "category": "BeReal",
         "notes": "https://djangofaiola.blogspot.com",
-        "paths": ('*/mobile/Containers/Data/Application/*/Library/Caches/disk-bereal-ProfileRepository/*'),
+        "paths": ('*/mobile/Containers/Data/Application/*/Library/Caches/disk-bereal-ProfileRepository/*',
+                  '*/mobile/Containers/Data/Application/*/Library/Caches/AlexisBarreyat.BeReal/Cache.db*',
+                  '*/mobile/Containers/Shared/AppGroup/*/EntitiesStore.sqlite*'),
         "output_types": [ "lava", "html", "tsv", "timeline" ],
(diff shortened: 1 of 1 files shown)


Johann-PLW:
Thank you for updating these two modules.
**For BeReal module**:
I noticed that all records were doubled for `Accounts` and `Friends` artifacts. Do they come from a different source file or is it a little bug? Do you have time to have a look?

**For Waze module**:
We can push the update as it is.
However, I think we can replace the actual output of `track_gps_quality` with "all" to add kml export as there are latitude and longitude data with a timestamp.
In a near future, we could also split the coordinates column of `Favorite locations`, `Recent locations`, `Searched locations` and `Session info` artifacts in two separate latitude and longitude columns to add a kml export.

Thank you very much for your contribution.

djangofaiola:
**For BeReal module:**
Double/triple records are present in almost all artefacts. Different source files: Cache.db, EntitiesStore.sqlite and various json.
The "Source file name" column contains the list of files.
Can you help me understand why in the html report the column ‘Source file name’ even though I have formatted the cell with `<br>` does not display the line break? Where am I going wrong?

**For Waze module:**
I forgot to put the output to “kml” in the update. I replaced the output of track_gps_quality with “all”.
I have already split the coordinate columns into two columns latitude and longitude in the favorite places, recent places, searched places and sessions info artifacts, but I forgot to include the "kml" output.
For the "session info" artifact I have two coordinates with column names other than "Latitude" and "Longitude" and I did not add the output to "kml".
Done!

Grazie a Voi

Johann-PLW:
Hi @djangofaiola,

If you're inserting HTML tags for better rendering in the HTML report, you need to create the `html_columns` key in the `__artifact_v2__` block and specify the column names.

This was documented in the `artifact_info_block.md` file on the repo in `admin/docs` but an update last week accidentally removed this part:
> - `html_columns`: The name of the columns that contain HTML code to render it properly in the HTML report. This key is used to populate the optional argument `html_no_escape` in the report.write_artifact_data_table() function.

You have an example in `splitwiseGroups` and `splitwiseNotifications` notification from `splitwise.py` module.

Merci beaucoup.

JamesHabben:
@Johann-PLW did we update the html output engine to take a list of source files? that would be a way better option than injecting formatting tags into the html.

djangofaiola:
Hi @Johann-PLW ,
Of course, I put the column names in the 'html_columns' key.
My data is formatted in HTML, but the HTML report is missing the tags!
It might be a small bug, or I could be doing something wrong.
I would appreciate it if you could take a look at my code.
Thanks!

djangofaiola:
@JamesHabben ,
Having the ability to input a list of source files is useful for most situations. In some cases, it might become too long.
In the analysis of BeReal's 'Cache.db', each record could refer to an external file in the fsCachedData subfolder, and the list could become very lengthy.

djangofaiola:
@Johann-PLW , @JamesHabben Thank you for the great work done

JamesHabben:
@djangofaiola ibthink I misunderstood then. In other modules where the source files can be a long list and unique for each record, we have gone to adding a `source_file` column to the table. The artifact level source file value can point to the database where the module is extracting information, and then the record level source file value can point to its specific file path(s) that were interpreted or extracted. 

We also have a new media manager library that we are moving modules over to when working with media type files. This will help our new LAVA viewer to enrich these for better examiner review. I forget the module that @Johann-PLW started with though , maybe WhatsApp?

If you are curious, look at the `check_in_media` function. We are still working on developer documentation for this.

JamesHabben:
take a look at an example of what i am describing in my last comment. there is a column added on the end of the records to show the plist file where this wifi info came from. many modules do this in case they are extracting data from multiple files in different paths or names.
https://github.com/abrignoni/iLEAPP/blob/b14e3aeab6f3c541957ed5bd07460c9159a01270/scripts/artifacts/appleWifiPlist.py#L218-L267

it ends up in the output like this:
image

JamesHabben:
also @djangofaiola , some new documentation about the media management api is up. open to feedback and questions.

https://github.com/abrignoni/iLEAPP/wiki/Artifact-Module-API#check_in_media

JamesHabben:
@djangofaiola i understand better now what you are describing. thanks for sticking with us here!

it looks like a bug in the artifact processor core having to do with a combo of media and overloaded data sets for html output.

we split the data sets into test and html types here
https://github.com/abrignoni/iLEAPP/blob/92e095f60eec073203f16b1586ddf5825a59b0d1/scripts/ilapfuncs.py#L340-L341

there a few lines down we check for media types. the `get_data_list_with_media(...)` func then does another data set split but its based on the text only data set. effectively, we are discarding the previously split out html data set with a new text only set.
https://github.com/abrignoni/iLEAPP/blob/92e095f60eec073203f16b1586ddf5825a59b0d1/scripts/ilapfuncs.py#L351-L354

only happening because of the media check in. :) i will put together an issue for this so we can work on a fix. want to hold on this PR until that is resolved or should we push through?

JamesHabben:
btw, @djangofaiola looks like one of your commits raised a couple conflicts in the contacts artifact. can you check that out in the meantime?

PR #1174, opened 2025-05-26
Update ilapfuncs.py - get_birthday fix

kobo220:
If a birthday is added to a contact without a year, it is stored with the year 1604. These dates can cause an issue in Windows Python with the current function that is used to parse these dates. If one of these dates is found the addressBook.py artifact causes the error: "OSError: [Errno 22] Invalid argument" and completely fails to output any results. This change replaces datetime.utcfromtimestamp (which is also depreciated) with manually creating a Cocoa Epoch to avoid out of range date problems. The datetime.fromtimestamp function is also susceptible to this problem.

### DIFF
--- a/scripts/ilapfuncs.py
+++ b/scripts/ilapfuncs.py
@@ -1121,8 +1121,8 @@ def convert_plist_date_to_utc(plist_date):
         return plist_date
 
 def get_birthdate(date):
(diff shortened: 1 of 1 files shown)


Johann-PLW:
Thank you very much @kobo220 for updating the function and solving this issue.

PR #1190, opened 2025-06-04
add 'file' run type

JamesHabben:
- added new `FileSeekerFile` class to run against a single file input
- added `-t file` command line argument
- added extra validations against path parameters
- added wildcard to `logarchive` search path


abrignoni:
I'm good to me. I will defer to @Johann-PLW for merging after his review.

Johann-PLW:
@JamesHabben 
It's also good for me.
I've just noted a doubled check about input file validation, once in ileapp.py line 51: `os.path.isfile(abs_input_path)` and again in the FileSeekerFile class line 371 `os.path.isfile(self.single_file_abs_path)`.

My only fear is the use of wildcard to logarchive search path. If several logarchive files belonging to different devices are present, they will all be parsed and all event messages will be merged.
I'm not sure that's what we're expecting.
But if @abrignoni is OK with that, I can merge this.

abrignoni:
Your thoughts @JamesHabben ?

JamesHabben:
I'll take another look for the file validation. The main one I wanted to add was the upfront check in `ileapp.py` since I ran into an issue with ileapp continuing on into processing a bad input.

For the multiple log files going into one:
1. This obviously won't happen with the single file input run type so not an issue
2. With the FFS/folder input, there is a chance of this happening though I don't think it will be very common. I think most folks will put a single log file in the folder with their extraction. 

That said, it's a possibility and you are right to call it out. Good catch!

The multiple log files will show in the source files path list, so there is indication of the combined data. Do we need to do something more to address it?

Johann-PLW:
Thank you @JamesHabben
I think I can merge as it is.
I will have a look later maybe to add a warning in the logs and in the HTML report that multiple logarchive files were parsed.

PR #1197, opened 2025-06-19
Chrome bookmarks modified

dumbermore:
This patch contains:

1. Added support for backup bookmark file (**Bookmarks.bak**)
2. Iterating through all bookmarks 

_this closes #937_

### DIFF
--- a/scripts/artifacts/chrome.py
+++ b/scripts/artifacts/chrome.py
@@ -96,6 +96,7 @@ __artifacts_v2__ = {
         "author": "@stark4n6",
         "version": "0.0.3",
         "date": "2024-11-10",
+        "last_update_date": "2025-06-20",
         "requirements": "none",
         "category": "Chromium",
         "notes": "",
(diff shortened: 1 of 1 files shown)


stark4n6:
@dumbermore going to merge, but will try to get this updated for LAVA support unless you wanted to give it a try

JamesHabben:
@stark4n6 i think we just merge this in. the whole module needs some refactoring for outputs as many artifacts are doing html and lava output manually. this is standalone improvement to the module. new issue can tackle the refactor of output types.

dumbermore:
As mentioned the entire module can benefit from a rewrite. Also the fact that _chrome.py_ handles all chromium based browsers is not obvious from a glance and it is missing some browser support like for _Edge_ which I originally intended to add in a later separate pr.

I have done some testing on my end where I found the _Bookmarks.bak_ issue after which it was able to iterate through all the bookmarks.

stark4n6:
@dumbermore @JamesHabben merging, and I'll see about updating things for LAVA and more friendly naming. I thought Edge was supported but I could be thinking about ALEAPP instead

PR #1199, opened 2025-06-20
Update health.py

MetadataForensics:
Parses Apple Health Wrist Temperature, as outlined within Health Application > Summary > Show All Health Data > Wrist Temperature > Show All Data > All Recorded Data

### DIFF
--- a/scripts/artifacts/health.py
+++ b/scripts/artifacts/health.py
@@ -191,6 +191,19 @@ __artifacts_v2__ = {
         "paths": ('*Health/healthdb.sqlite*',),
         "output_types": "standard",
         "artifact_icon": 'smartphone'
+    },
+    "healthWristTemperature": {
+        "name": "Health - Wrist Temperature",
+        "description": "Parses Apple Health Wrist Temperature, as outlined within Health Application > Summary > Show All Health Data > Wrist Temperature > Show All Data > All Recorded Data",
+        "author": "@SQLMcGee for Metadata Forensics, LLC",
+        "version": "0.1",
+        "date": "2025-06-20",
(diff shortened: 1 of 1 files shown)


stark4n6:
@MetadataForensics do you have test data for this? I haven't been able to find a sample extraction to check this out

MetadataForensics:
@stark4n6 I apologize, I was out of the office yesterday. My dataset for this is 18.5 as well so I do not believe it's a version issue. To clarify, you pulled an extraction of your phone that has wrist temperature information recorded from an Apple Watch and ran it through iLEAPP with this artifact? Or did you run the query against the database in DB Browser or another database viewer? It can also work within DB Browser but you need to attach the healthdb.sqlite database as well as it pulls data from there too.

MetadataForensics:
@stark4n6 Interesting, curious to see what you find. The iLEAPP artifact works on all datasets I have which include Wrist Temp, or give the "No data found for Health - Wrist Temperature" if the data_type is not present.

MetadataForensics:
@stark4n6 The amount of SQLite databases you accumulate in this work is something else, no worries - glad the query works on your end too. I'd thought about using the builds_id with this but hadn't included it in the attempt to duplicate the exact info within Health Application > Summary > Show All Health Data > Wrist Temperature > Show All Data > All Recorded Data. An example is shown below.. I can see both points: 1) data replication as shown through the app 2) usefulness to the end user.. What are your thoughts?

IMG_2121
IMG_2122

stark4n6:
@MetadataForensics will merge for now, we can always change it quick later if necessary, thanks again for the contribution!

PR #1203, opened 2025-06-23
Media mgr updates and context class

JamesHabben:
- create context class to hold more info about current module and artifact
- update media functions to use info from context class to reduce params required from modules
- see commit notes for more details


Johann-PLW:
Everything is OK.
Thank you for this welcome improvement.

PR #1221, opened 2025-07-28
User defaults

snoop168:
Parses plists containing the userdefaults stored by applications


stark4n6:
@snoop168 was this working? saw the linter failed on imports, not sure if biplist is in the requirements

PR #1224, opened 2025-08-01
Modified LinkedIn Artifacts and added Airbnb message artifact

kalink0:
Modified LinkedIn Artifacts to be LAVA compliant
Added Aribnb artifact (message) - more to come later

### DIFF
--- a/scripts/artifacts/Airbnb.py
+++ b/scripts/artifacts/Airbnb.py
@@ -1,0 +1,72 @@
+# Airbnb App (com.airbnb.app)
+# Author:  Marco Neumann (kalinko@be-binary.de)
+# Version: 0.0.1
+# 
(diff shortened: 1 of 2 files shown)


JamesHabben:
thanks for catching those linter errors. its a new check we put in and are still working on tuning it. open to feedback if you have experience around that. i haven't reviewed code, just noticed commenting on the linter fixes.

we also have a new artifact processor function using a single parameter context object rather than the multiple parameters. not documented yet, but you can see some examples that have been converted if you are curious.
- https://github.com/abrignoni/iLEAPP/blob/main/scripts/artifacts/addressBook.py
- https://github.com/abrignoni/iLEAPP/blob/main/scripts/artifacts/cachev0.py

PR #1225, opened 2025-08-04
add support for XBApplicationSnapshotManifest metadata from applicationState.db

mxkrt:
Hello,

I've created a plugin that adds support for parsing the metadata stored in the XBApplicationSnapshotManifest blobs in applicationState.db. Hopefully this plugin can be added to the main project. Please let me know if any more steps are required before creating a pull request.

Kind regards,

mxkrt


mxkrt:
Not sure if it was the proper thing to do, but I've justed synced my fork with the main branch.

mxkrt:
I see the message "1 workflow awaiting approval" in this PR. Is this something I can/should do?

stark4n6:
@abrignoni what are your thoughts? It might be just as simple as adding the extra parsers as extra functions and not replacing unless the parsers are parsing out similar data.

JamesHabben:
ideally, we group similarly focused artifacts together in a single module. thats the purpose of the artifact block structure so each artifact can be a standalone output view, even with distinct credit to authors. the other consideration would be if an artifact takes a significant amount of time to process its data. we currently dont have an ability to deselect artifacts within a module - its run the module as a whole or not.

stark4n6:
@mxkrt looks good, the only thing I may ask is do we want to have more friendly header column names? (without underscores and camel casing?)

JamesHabben:
The given column names get converted by the core during processing, so these should be plain English display headers. 

@stark4n6  probably should be noted somewhere in the docs as it's counterintuitive to developers

mxkrt:
Thanks for merging the PR and for all the feedback and the link to the testing setup documentation!

PR #1238, opened 2025-08-26
Improve media file lookup speed

JamesHabben:
Replaces the slow, linear search in `check_in_media` with a high-performance, two-stage lookup.

The previous O(N) search caused severe slowdowns on artifacts with many files. This is now an O(k) operation using a lazy-loaded dictionary on the `Context` object.

A new `Context.get_source_file_path()` API is introduced to centralize this new, efficient logic.

### DIFF
--- a/scripts/context.py
+++ b/scripts/context.py
@@ -1,3 +1,7 @@
+from os.path import basename
+from pathlib import Path
+
+
 class Context:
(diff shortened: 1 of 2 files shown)


abrignoni:
This needs to be ported to the other LEAPPs.

PR #1255, opened 2025-09-14
knowledgeC artifacts added

mxkrt:
I've added some common artifacts to the knowledgeC analysis module related to App Usage, Lock state and the state of the screen backlight. Hopefully this can be added to the project.


prosch88:
Add an underscore before the arguments. This shows that the arguments are not used in this function, and Pylint is satisfied.

mxkrt:
Thanks, good to know! Now fixed.

mxkrt:
Thanks! For me it is fine if we leave the code as is for this PR, but I will certainly keep the new approach in mind for my next contribution.

PR #1279, opened 2025-09-29
bug: update lava conversation output

### LINKED ISSUE #1277: bug: update alternate data views
LAVA has been updated to use more appropriate terms for the alternate data view previously called `chat` and `thread`. the `lavafuncs` module was hardcoded to look for `chat` and is further hardcoded to insert the field names.
https://github.com/abrignoni/iLEAPP/blob/a6f624fc297be7fd3e9534747158e0d4a5a61299/scripts/lavafuncs.py#L99-L115

### fixes
1) the function should be adjusted to be more pass through. in other words, it should iterate the alternate data views in the artifact and pass through any that were declared. this will allow a new alternate view to be added in LAVA and in artifact blocks without having to specifically add that view to this function.
2) it should also pass through any field names that we can to minimize further maintenance issues in the future as we expand or adjust these views.
3) keep in place the column sanitization since we have module developers use plain english data headers and do the mapping in the background for sql safe column names

note: LAVA currently has a conversion in place to identify any `chat` data views and any `thread` field names. that will stay in place for a few version until we get artifacts converted over. 

related to https://github.com/abrignoni/iLEAPP/issues/1254

JamesHabben:
- update output to convert old chat to new conversation naming
- update to dynamically detect column names that need sql sanitizing

closes #1277

### DIFF
--- a/scripts/lavafuncs.py
+++ b/scripts/lavafuncs.py
@@ -97,31 +97,41 @@ def lava_process_artifact(category, module_name, artifact_name, data, record_cou
         artifact["object_columns"] = [{"name": name, "type": type_} for name, type_ in object_columns.items()]
 
     if data_views:
-        if chat_params := data_views.get("chat"):
-            sanitized_params = {}
+        view_params = None
+        view_name = None
 
-            #Boolean value is whether or not to sanitize the column name. Should do this for parameters that map to columns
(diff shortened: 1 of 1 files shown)


stark4n6:
@JamesHabben @prosch88 merged

PR #1280, opened 2025-09-30
Session

snoop168:
Including initial keychain work. Place GK keychain in same folder as extraction, or it will attempt to locate the UFED keychain inside the zip.

### DIFF
--- a/ileapp.py
+++ b/ileapp.py
@@ -316,17 +316,8 @@ def crunch_artifacts(
     logfunc('By: Alexis Brignoni | @AlexisBrignoni | abrignoni.com')
     logfunc('By: Yogesh Khatri   | @SwiftForensics | swiftforensics.com\n')
 
-    logfunc('Checking for keychain file in the same folder as the input "*keychain.plist\n')
-    input_path_directory = Path(input_path)
-    if extracttype != 'fs':
-        input_path_directory = input_path_directory.parent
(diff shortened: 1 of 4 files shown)


snoop168:
can someone else review this? I think the error @stark4n6 was getting might be specific to his environment.

abrignoni:
Thanks for this substantial work — sqlcipher support + keychain handling is valuable. A few things before it can move forward: it's ~9 months old now and `main` has changed a lot (this touches core `ilapfuncs.py`, adds dependencies, and CI is currently failing), and it overlaps with the keychain decryption work in #1543. Could you rebase onto the latest `main` and get CI green? This also needs an owner-level decision on the sqlcipher/keychain infrastructure direction given the overlap with #1543. Happy to re-review once it's rebased.

abrignoni:
Thank you for this, and for pushing the SQLCipher and keychain direction early. That direction is what iLEAPP ended up taking: the pure-python SQLCipher reader (scripts/sqlcipher_decrypt.py) plus the keychain flow (scripts/ios_keychain.py, --keychain / GUI field) now provide this in core, without the compiled sqlcipher3 dependency. Session itself is now covered on that infrastructure as a leaf artifact in #1928 (messages + contacts, validated by decrypting the real databases in our iOS test images). Since the core plumbing and the Session parser both landed that way, and this branch is ~11 months behind main with failing CI, I am closing it as superseded. The early groundwork here is genuinely appreciated.

PR #1287, opened 2025-10-03
Maite modules batch 1

Maite2003:


Maite2003:
> in data headers list, if you see any dates being stored, we should designated them with the `datetime` render type. the airbnb has a few in the header list. https://github.com/abrignoni/iLEAPP/blob/3946e10b4681c3d807f04d1daf4b006eae55ae26/scripts/artifacts/Airbnb.py#L64C14-L64C35
> 
> for reference: https://github.com/JamesHabben/iLEAPP/blob/main/admin/docs/module_updates.md#4-adjust-the-main-function

@JamesHabben For the datetime render type in the data headers list, should we use it even if the dates are stored as strings? I was under the impression that datetime was only for actual datetime objects.

JamesHabben:
for now, the datetime storage in sqlite only accepts unix seconds. we are looking to change this to accept sub-second resolution but the LEAPP framework hasnt been adjusted for that yet. we can round off to the second for now, but we can then start an issue to track a list of modules that need updates when we get the sub-second resolution in place.

PR #1289, opened 2025-10-04
Setup this repo as a proper Python project with a pyproject.toml file and a main package

Iriamu19:
Hi,

I don't know if this is something you are interesting in, but I'm working on improving the organization of this repo to setup it as a proper Python project with a pyproject.toml and a main package with modules.

When done, this will allow iLEAPP to be built as a proper wheel and to be pushed to PyPi.

It will also be easier to install using `pip install ileapp` or `pipx install ileapp`.

For now, I created a main package named ileapp/ at the root of the repo and put scripts/, assets/, ileapp.py and ileappGUI.py inside.

I'm managing the project with uv.

I'm working on Linux and it seems to work well. I can build a wheel and install ileapp from it. I'll try to test it on macOS and Windows (platform specific wheels could cause some issues). Some work is still needed for the pypinstaller *.spec files.

What do you thinks about this?

Best regards.


JamesHabben:
no response. closing.

PR #1294, opened 2025-10-07
Update os_builds.json

Johann-PLW:
Add:
- iOS 18.7, iOS 18.7.1, iOS 26.1 beta 2
- macOS 14.8.2 RC, macOS 15.7.2 RC, macOS 26.1 beta 2
- visionOS 26.1 beta 2
- watchOS 26.1 beta 2
- tvOS 26.1 beta 2
- Windows 11 version 25H2

Add missing iOS versions:
- iOS 15.8.4, iOS 15.8.5
- iOS 16.7.11, iOS 16.7.12
- iOS 17.7.6, iOS 17.7.7, iOS 17.7.8, iOS 17.7.9, iOS 17.7.10
- iOS 26.0

### DIFF
--- a/scripts/data/os_builds.json
+++ b/scripts/data/os_builds.json
@@ -651,6 +651,8 @@
         "19H380": "iOS 15.8.1",
         "19H384": "iOS 15.8.2",
         "19H386": "iOS 15.8.3",
+        "19H390": "iOS 15.8.4",
+        "19H394": "iOS 15.8.5",
         "20A5283p": "iOS 16.0 beta 1",
         "20A5303i": "iOS 16.0 beta 2",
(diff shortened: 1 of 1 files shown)


Johann-PLW:
Yes, these new JSON files replace OS_build, device_id and code_name of build_ids.py file.
Pylint complained that this file was too big.
We have redesigned the OS_build dictionary to use nested dictionaries with the os family as the outer key, and the build as inner. then version as values. So there are no duplicate keys anymore.
JSON files are loaded once when a module need them and they can be accessed from an artifact without any specific import.

I will clean up the build_ids.py file once I have finished updating the modules that imported the build_ids.py file to lookup OS versions or model names, just to not break averything.

PR #1301, opened 2025-10-08
Maite modules batch 2

Maite2003:
Refactoring: Timestamps and Argument Handling

This PR introduces two primary updates:
1.  **Timestamp Correction:** I removed the database conversion function (`datetime(time/1000,'unixepoch')`) from the query. Instead, we now pull the raw Unix timestamp and use `convert_unix_ts_to_utc()` in Python. Is this the approach we should use?
2.  **Artifact Arguments:** Updated artifact function signatures to accept a context object instead of individual parameters.


Maite2003:
i don' have any test data for AMDSQLiteDB

stark4n6:
@JamesHabben let me look, I think I made this parser so I think I actually had test files (either on my own test devices or from a sample)

JamesHabben:
Most of these changes are pretty straight forward. The only parts that concern me are the dates to ensure we are using the correct function based on the source data. Once that is validated, I think this PR is good to merge.

PR #1315, opened 2025-10-24
Maite modules batch 5

Maite2003:
- divided apple passes artifact into PK passes and Nano
- added validation to ilapfuncs get_plist_content and get_plist_file_content to check if dict before doing get
- update artifacts to use context
- on applicationStateDB i tried replacing _parse_blob with a plist utility function but couldn't find a way to do it.

Pk passes, WorldClock, applicationStateDB and wifiPlist are tested. I couldn't find data on transactions or Nano passes


Maite2003:
@JamesHabben I solved the lint error. Also, I've noticed this plist parsing logic is repeated in several files. Do you think it would be a good idea to extract it into a function in ilapfuncs? That way, we can handle any parsing issues or logic changes in a single place instead of updating every artifact

JamesHabben:
@Maite2003 can you let us know if you created an issue for the plist code? If so please link here. Otherwise we need to get one created so we can clear this PR.

PR #1317, opened 2025-10-28
Maite modules batch 6

Maite2003:
Refactor AWESearch and audiTrip data to v2 with Context,
Deleted unused imports and changed source path in appSnapshots for consistency with other artifacts,
Updated ashHistory and atxDatastore to use Context
For atxDatastore I also updated the db query to get the raw data and then use functions to get the timestamp in the correct format

appSnapshots and atxDataStore are tested, for the rest I couldn't find test data

### DIFF
--- a/scripts/artifacts/appSnapshots.py
+++ b/scripts/artifacts/appSnapshots.py
@@ -22,7 +22,7 @@ __artifacts_v2__ = {
 from pathlib import Path
 
 from PIL import Image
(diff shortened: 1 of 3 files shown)


JamesHabben:
- theres a few linter errors that could be cleared up easy enough. mostly tab misalignments. 
- the `liblzfse.error` error seems we can just remove that third error type since that error doesn't exist.

can you tackle these @Maite2003 ? otherwise, it looks good to me.

PR #1320, opened 2025-10-31
feature: update media checkin to fix path bug and allow modules to set more media info

JamesHabben:
- add `is_embedded` to media item table
- remove `media_path` from media references table
- create central media folder and link/copy all media
- create html media folder and link/copy all html media
- change sql inserts to use parameters
- move media check in logic to single internal function
- public media check functions point to internal
- add output parameters to context
- add force params to let modules set mime type, extension, created, modified


JamesHabben:
new commit to resolve the db source path issue you mentioned to me. ready for review when you have a chance @Johann-PLW

PR #1323, opened 2025-11-03
Maite modules batch 8

Maite2003:
refactor biomes artifact to v2. 
- Most of them were already using the v2 dict but with version and date, so i just replaced that with creation_date and last_update_date and changed the artifact params to context.
- Removed timezone conversions
- Some of them also had a typess dict defined in a single line which was difficult to read. I reformatted it into a multi-line dict
- Changed the output source file to 'see Filename for more info' so it's consistent with other artifacts
- Intents and Notes still had manual report generation. I replaced it by returning the necessary values. 
- In Useractmeta, Locationactivity and Intents, replaced deserialize_plist_from_string with get_plist_content

Backlight, DeviceSync, Notifications, NowPlaying, TextInputses, UserActMeta and Intents are tested and working fine. For the rest i couldn't find any test data

### DIFF
--- a/scripts/artifacts/biomeAirpMode.py
+++ b/scripts/artifacts/biomeAirpMode.py
@@ -15,7 +15,7 @@ __artifacts_v2__ = {
 
 import os
 import blackboxprotobuf
-from datetime import *
+from datetime import timezone
 from scripts.ccl_segb.ccl_segb import read_segb_file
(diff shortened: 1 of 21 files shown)


JamesHabben:
the only lint error left is a todo. I am not fixing that. kick that over to @snoop168

PR #1327, opened 2025-11-05
Maite modules batch 9

Maite2003:
Refactor artifacts to v2 and to use Context.
Also added plist validation 

- In bluetoothPairedReg.py changed lastseen from a string to a datetime object in utc. Also marked the data header column
- In booking.py removed the preferences artifact since it was only used to set a global variable which was only used by the booking_booked artifact. Instead, I added the path to that artifact and set the variable inside the artifact to ensure it is set before the artifact is executed
- In booking.py fixed some of the HTML data_lists. Also, in the booking_notifications artifact, changed the query to retrieve raw data and the perform the conversion with convert_cocoa_core_data_ts_to_utc function

I tested blockedContacts, bluettothOther, bluetoothPairedLE and bluetoothPairedReg. For booking i couldn't find test data


JamesHabben:
- the `booking_account` artifact was appending 19 column values when only 18 were expected. removed a double source file value
- the `booking_preferences` function was removed, but the artifact header was still in place. removed entry
- the `bluetoothPairedReg` module was returning inside the loop. moved to outside the loop.

JamesHabben:
after committing, GitHub was showing issues with booking module. turns out @djangofaiola updated this a few months back. removed the booking from this pr since its no longer needed.

PR #1329, opened 2025-11-06
Fix minor documentation typo in ileappGUI.py

anastanira:


JamesHabben:
Hi @anastanira, was this PR intended to be opened here? there appears to be customization to specific organization needs which are not suited for the general use of iLEAPP.

PR #1344, opened 2025-11-21
Batch #2 changes to v2

FranciscoVeronING:
Update:
**walStrings.py** : Tested with iOS15 image. No issues **voiceTriggers.py** : Tested with iOS 15 image. No issues. **voiceRecordings** and ** vippsContacts** : no data to test ** voiceemail.py** : Tested with iOS 15 image. No issues


JamesHabben:
it looks like there is a date in the voiceTriggers module that is not getting passed as a date object with the header tagging. otherwise i think this looks good, but @stark4n6 have a look once that date is addressed?

edit: there are some linter errors to address too

JamesHabben:
this one took some work to refactor a bit. the walstrings module makes text files with the found strings and the html output has links to open the text files for review. that's not a good review path for lava output, so refactored the code to build a summary and detail. html output is nearly the same as it was - producing the text files - only now the strings are deduplicated before writing into the file. for lava, the deduplicated strings are written to the lava db each string getting its own row. running this against josh's ios15 image produces almost 1 million rows. i also changed the string search pattern since the previous one was getting a lot of white space and control characters included which was creating a lot of empty lines and less-than-four-character string matching.

here is a sample of the detail artifact in lava
image

here is summary
image

@stark4n6 @abrignoni thoughts on this change?

stark4n6:
@JamesHabben looks good

PR #1358, opened 2025-12-01
notes: add small guards to avoid OSError when processing attachments

nilamwinarti13:
Add existence check and try/except around imghdr.what() and thumbnail creation to prevent plugin crash on unsupported or corrupted files. This is a minimal change intended for a hotfix; behavior unchanged for normal images.

What:
- Add minimal defensive checks in notes artifact to prevent OSError when checking or opening attachment files.
  * Check file existence before running imghdr.what()
  * Wrap imghdr.what() in try/except and fallback to None on failure
  * Wrap thumbnail writing in try/except to prevent crash

Why:
- Certain platforms / datasets contain attachment paths (or formats like HEIC) that cause imghdr or PIL to raise OSError, which crashes the plugin.
- This is a small, non-invasive change to make the artifact robust while keeping behavior unchanged for normal image files.

### DIFF
--- a/scripts/artifacts/notes.py
+++ b/scripts/artifacts/notes.py
@@ -7,6 +7,7 @@ from PIL import Image
 import imghdr
 import zlib
 import binascii
+import os
 
 from scripts.artifact_report import ArtifactHtmlReport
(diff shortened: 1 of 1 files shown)


JamesHabben:
the remaining linter errors are things to address in a separate modernization effort. this code is good to merge now.

PR #1361, opened 2025-12-02
BugFix error AI Gen in IOS 14, Album Records NAD and some column not found

AkagaRianma:
[Diff before and after.docx](https://github.com/user-attachments/files/23880831/Diff.before.and.after.docx)

dataset: https://digitalcorpora.s3.amazonaws.com/corpora/scenarios/magnet/2021%20CTF%20-%20iOS.zip

I want to fix Ph17GenAIDetected.py so that it doesn't show error when parsing on IOS < 18. I also want to fix Ph20AlbumsNAD.py so that it can parse on IOS 14. I also added some fixes for missing columns on IOS 14 when parsing sqlite databases. I prioritize the fixes in Ph17GenAIDetected.py and Ph20AlbumsNAD.py files.


JamesHabben:
Closing this as superseded/conflicting. This PR has many files together which makes conflicts and overlaps even more likely. The MapsSync, ParseCD Cache, and Health fixes have already landed in newer PRs, and the remaining Photos column-guard idea should be handled separately from current main using the existing does_column_exist_in_db helper. This PR now contains stale overlap and lint issues, so it is not safe to merge as-is. Photo stuff might possibly be addressed in #1499 but if not, then this should get looked at and opened into a new focused PR.

abrignoni:
Follow-up: the error classes reported here (Gen AI parsing on iOS 14, Album Records NAD, column-not-found) are addressed on `main` as of #1713 — the reworked Photos.sqlite parsers version-gate the Gen AI queries (iOS 18+) and add iOS 16+ column-existence guards for unmigrated databases. Verified with full runs on iOS 17.5.1 and 18.3.2 images with zero artifact errors. Thanks @AkagaRianma for the report and initial fix work.

PR #1363, opened 2025-12-04
Fix: FitnessWorkoutsLocationsData plugin 'no such table'

KemalRajasa:
**Plugin edited:**

`scripts/artifacts/FitnessWorkoutsLocationData.py`

**Issue:**

On iOS 15.0.2 that doesn't have fitness app, parsing its database will give an error:
```
[167/446] Fitness [FitnessWorkoutsLocationData] artifact started
Reading Fitness artifact had errors!
Error was no such table: associations
Exception Traceback: Traceback (most recent call last): File "/home/rjskali/forensic/forensic-tools/iLEAPP/ileapp.py", line 491, in crunch_artifacts plugin.method(files_found, category_folder, seeker, wrap_text, time_offset) File "/home/rjskali/forensic/forensic-tools/iLEAPP/scripts/artifacts/FitnessWorkoutsLocationData.py", line 35, in get_Health cursor.execute(''' sqlite3.OperationalError: no such table: associations
```

**Fix Implemented:**

Implemented a try-except block to handle if there were no fitness app or `associations` table mentioned in error log earlier

### DIFF
--- a/scripts/artifacts/FitnessWorkoutsLocationData.py
+++ b/scripts/artifacts/FitnessWorkoutsLocationData.py
@@ -32,8 +32,8 @@ def get_Health(files_found, report_folder, seeker, wrap_text, timezone_offset):
     
     # Fitness Workouts Location Data Analysis
         try:
-        cursor.execute('''
-        SELECT
+            cursor.execute('''
+            SELECT
             datetime(workout_activities.start_date + 978307200, 'UNIXEPOCH') as "Workout Start Time",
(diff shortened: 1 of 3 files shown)


JamesHabben:
the `except` block as it was captured all errors from the sqlite query and the message being output assumed the error was generated by the missing `associations` table only. I made the exception a general exception which will write out the error that was generated. then above the query added a specific check to see if the table exists so we can then write a correct error based on the test.

PR #1364, opened 2025-12-04
implement: maps sync schema if 'column not found'

KemalRajasa:
**plugin modified**

`scripts/artifacts/mapsSync.py`

**Issue:**

Some column not found in this error log example is ZLONGITUDE

```
[251/446] mapsSync [mapsSync] artifact started
Reading mapsSync artifact had errors!
Error was no such column: ZHISTORYITEM.ZLATITUDE1
Exception Traceback: Traceback (most recent call last): File "/home/rjskali/forensic/forensic-tools/iLEAPP/ileapp.py", line 491, in crunch_artifacts plugin.method(files_found, category_folder, seeker, wrap_text, time_offset) File "/home/rjskali/forensic/forensic-tools/iLEAPP/scripts/artifacts/mapsSync.py", line 44, in get_mapsSync cursor.execute(''' sqlite3.OperationalError: no such column: ZHISTORYITEM.ZLATITUDE1
```

**Fix:**

Implementing nested try-except to iterate through databases column name, it will use the last except block if there are no column name match. The last except block will use empty string _NULL_ to replace the value of column that had no match with the target iOS image

### DIFF
--- a/scripts/artifacts/mapsSync.py
+++ b/scripts/artifacts/mapsSync.py
@@ -1,9 +1,15 @@
 import sqlite3
-import textwrap
 import blackboxprotobuf
 
 from scripts.artifact_report import ArtifactHtmlReport
-from scripts.ilapfuncs import logfunc, tsv, kmlgen, timeline, is_platform_windows, get_next_unused_name, open_sqlite_db_readonly
+from scripts.ilapfuncs import (
+    logfunc,
+    tsv,
+    kmlgen,
+    timeline,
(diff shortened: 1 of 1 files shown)


JamesHabben:
the `except` statements were assuming any error raised would be the missing column error. I refactored to specifically check for the potentially missing columns and use a reduced query based on those tests, and let the except print out the actual error if there is one. remaining linter errors can get cleaned up when the module get modernized.

PR #1365, opened 2025-12-04
Implement: safe query for column mismatch in photosDbexif 

KemalRajasa:
**Plugin modified:**
scripts/artifacts/photosDbexif

**Issue**

On older iOS version image there is a chance that a column might missing or the entire table doesnt exist such as this error message i encountered

```
[289/446] photosDbexif [photosDbexif] artifact started
Reading photosDbexif artifact had errors!
Error was no such column: ZADDITIONALASSETATTRIBUTES.ZCREATORBUNDLEID
Exception Traceback: Traceback (most recent call last): File "/home/rjskali/forensic/forensic-tools/iLEAPP/ileapp.py", line 491, in crunch_artifacts plugin.method(files_found, category_folder, seeker, wrap_text, time_offset) File "/home/rjskali/forensic/forensic-tools/iLEAPP/scripts/artifacts/photosDbexif.py", line 69, in get_photosDbexif cursor.execute(''' sqlite3.OperationalError: no such column: ZADDITIONALASSETATTRIBUTES.ZCREATORBUNDLEID
```

**Solution**

Implementend try-except method to try the query as the script intended and added the except block so the script wont return error, instead it just return the ZASSET table content

### DIFF
--- a/scripts/artifacts/photosDbexif.py
+++ b/scripts/artifacts/photosDbexif.py
@@ -1,4 +1,5 @@
 import os
+import sqlite3
 from datetime import datetime
 import pytz
 import json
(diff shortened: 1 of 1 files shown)


JamesHabben:
checking for columns and not assuming error is based on missing column. more modernization still needed.

PR #1371, opened 2025-12-06
fix: handle invalid Windows paths

hamasfaa:
### Cause
Notes attachments fail to load on Windows due to mixed path separators
(`/` + `\`) which produce invalid long-path strings when processed by imghdr.

### Fix
- Normalize attachment paths using `normpath()`
- Skip unsupported HEIC files to avoid imghdr open errors

### Result
The Notes artifact no longer crashes on Windows when attachments contain mixed
separators or HEIC formats.


JamesHabben:
Closing as superseded. The Notes attachment handling issue was already addressed in #1358, including the file existence check and safer thumbnail handling. This PR now conflicts with current main and would reintroduce broader exception handling/trailing whitespace.

PR #1376, opened 2025-12-06
Fix: Add fallback for missing sync_identity column on older iOS

Udong0:
The Problem: Currently, the artifact forces the selection of the sync_identity column in the SQL query. On older datasets (such as the Magnet User Summit 2021 CTF - Eli McFeely scenario), the healthdb.sqlite database schema does not contain this column.
This results in the following crash during parsing:
```
Error with ...\healthdb.sqlite:
- no such column: sync_identity
No data found for Health - Source Devices
```

The Solution: I implemented a dynamic schema check using ```PRAGMA table_info``` before executing the main query.

The script now checks if ```sync_identity``` exists in the ```source_devices``` table.

If it exists (Newer iOS): It queries the column as normal.

If it is missing (Older iOS): It substitutes the column with an empty string (```'' AS sync_identity```) to prevent the SQLite OperationalError.

### DIFF
--- a/scripts/artifacts/health.py
+++ b/scripts/artifacts/health.py
@@ -1155,7 +1155,6 @@ def health_watch_by_sleep_period(context):
 
     return data_headers, data_list, data_source
 
-
 @artifact_processor
 def health_source_devices(context):
     """ See artifact description """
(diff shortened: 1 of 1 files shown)


JamesHabben:
Closing as superseded. The same Health sync_identity fallback was already merged in #1415 via commit 71a820a45addce4594f3cec280065d42f87dd3f2, and the merged version handles the optional sync columns using the existing helper pattern.

PR #1377, opened 2025-12-06
Fix: Handle AttributeError when parsing invalid User Defaults plists

Udong0:
The Problem: In ```scripts/artifacts/userDefaults.py```, when a plist file fails to load (raises ```InvalidFileException```), the ```plist``` variable is assigned the string ```'INVALID FILE'```. The code then attempts to iterate over ```plist.items()```, causing an ```AttributeError: 'str' object has no attribute 'items'``` crash.

The Solution: I added a type check if ```isinstance(plist, dict):``` before the iteration loop.

If the file is invalid or corrupt, the loop is skipped gracefully.

The tool continues processing other artifacts instead of crashing the entire module.

### DIFF
--- a/scripts/artifacts/health.py
+++ b/scripts/artifacts/health.py
@@ -1155,7 +1155,6 @@ def health_watch_by_sleep_period(context):
 
     return data_headers, data_list, data_source
 
-
 @artifact_processor
 def health_source_devices(context):
     """ See artifact description """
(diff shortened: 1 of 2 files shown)


JamesHabben:
Closing as superseded. The User Defaults plist parsing issue was already handled in #1360 using the shared get_plist_file_content helper and dict guards. This PR also carries stale Health changes from #1376, so it is conflicted and no longer needed.

PR #1381, opened 2025-12-07
Fix IndexError in health_workouts by restructuring database record handling

hamasfaa:
## Fix tuple index out of range error in health_workouts artifact

Fixed IndexError caused by mismatched column count between two query branches when workout_activities table doesn't exist.

**Before:** The `else` branch query returned 26 columns but the code tried to access `record[26]`, causing an index out of range error. The missing "Location Type" column caused all subsequent indices to be off by one.

**After:** Split the record processing into separate branches matching the query logic. The `else` branch now correctly uses `record[25]` for added_timestamp, `record[21]` for hardware, and `record[15]` for temperature, with `None` inserted for the missing Location Type field.

### Error Message Before Fix

IndexError: tuple index out of range

**Files changed:**
- `scripts/artifacts/health.py`

### DIFF
--- a/scripts/artifacts/health.py
+++ b/scripts/artifacts/health.py
@@ -266,8 +266,6 @@ def health_workouts(context):
 
     data_list = []
 
-    celcius_temp = None
-
     activity_types = '''
         WHEN 1 THEN "AMERICAN FOOTBALL"
         WHEN 2 THEN "ARCHERY"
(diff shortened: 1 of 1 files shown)


JamesHabben:
I changed the approach to this fix. rather than duplicating the data read loop with the smaller query, we add a null value to the query so it's the same number of columns in both queries. then the data read loop can work the same without duplicate code.

great catch on finding this bug. thanks!

PR #1382, opened 2025-12-07
Fix: Handle missing 'ZLATITUDE1' column in MapsSync artifact for older iOS

Udong0:
The Problem: In ```scripts/artifacts/mapsSync.py```, the SQL query explicitly requests ```ZHISTORYITEM.ZLATITUDE1``` and ```ZHISTORYITEM.ZLONGITUDE1```. On older iOS versions, these columns do not exist in the ```ZHISTORYITEM``` table.
the crash: ```sqlite3.OperationalError: no such column: ZHISTORYITEM.ZLATITUDE1```
I updated the get_mapsSync function to perform a schema check using PRAGMA table_info before executing the main query.

### DIFF
--- a/scripts/artifacts/mapsSync.py
+++ b/scripts/artifacts/mapsSync.py
@@ -41,7 +41,18 @@ def get_mapsSync(files_found, report_folder, seeker, wrap_text, timezone_offset)
     
         db = open_sqlite_db_readonly(file_found)
         cursor = db.cursor()
-        cursor.execute('''
+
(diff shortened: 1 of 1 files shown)


JamesHabben:
Closing as superseded. The MapsSync legacy schema issue is already handled in #1364 using the project’s does_column_exist_in_db helper and broader schema-specific fallbacks. This PR uses raw PRAGMA checks and only covers part of the older-schema issue, so it is no longer needed.

PR #1383, opened 2025-12-07
fix: fix KeyError in wifiIdentifiers and using get() for optional fields

hamasfaa:
## Fix KeyError when IOMACAddress key is missing in wifiIdentifiers artifact

Fixed `KeyError: 'IOMACAddress'` that occurred when processing network interfaces without MAC address information.

**Before:** Code attempted to access `IOMACAddress` key directly without checking if it exists
**After:** Added check for IOMACAddress key existence. If the key is missing, sets MAC address field to empty string and continues processing. Also added safe access using `get()` for UserDefinedName and BSD Name

### Error Messages Before Fix
KeyError: 'IOMACAddress'

**Files changed:**
- `scripts/artifacts/wifiIdentifiers.py`

### DIFF
--- a/scripts/artifacts/wifiIdentifiers.py
+++ b/scripts/artifacts/wifiIdentifiers.py
@@ -42,27 +42,29 @@ def wifiIdentifiers(context):
         for key, value in pl.items():
             if key == 'Interfaces':
                 for y in value:
-                    try:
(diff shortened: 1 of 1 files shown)


JamesHabben:
resolved conflicts and adjusted the approach to the fix. the diff is showing extremely messy and not sure why, but code is the same.

PR #1385, opened 2025-12-08
fix: Fix base64 decoding errors and improve error handling

hamasfaa:
## Fix return order in appleMapsSearchHistory artifact 
Fixed multiple errors that prevented Apple Maps search history from being parsed correctly.

**Before:**
1. KeyError when accessing nested dictionary keys `protostuff['8']['2']['1']['201']['2']['2']['1']`
2. Base64 decode error with incorrect padding when processing malformed data
3. Wrong return order
4. Duplicate column names
5. Bytes objects in data causing JSON serialization errors

**After:**
1. Added safe dictionary access using `get()` with nested checks
2. Added try-except block in `shortbase64proto()`
3. Fixed return statement order
4. Renamed duplicate columns: 'Search Term' → 'Search Term 1' and 'Search Term 2', fixed 'Lat2' duplication to 'Lat3'
5. Converted `items` to string using `str(items)`

### Error Messages 
KeyError: '2'
binascii.Error: Incorrect padding
ValueError: too many values to unpack (expected 2)
sqlite3.OperationalError: duplicate column name: search_term
TypeError: Object of type bytes is not JSON serializable

**Files changed:** 
- `scripts/artifacts/appleMapsSearchHistory.py`

### DIFF
--- a/scripts/artifacts/appleMapsSearchHistory.py
+++ b/scripts/artifacts/appleMapsSearchHistory.py
@@ -8,36 +8,59 @@ __artifacts_v2__ = {
         "requirements": "none",
         "category": "Locations",
         "notes": "",
-        "paths": ('*/mobile/Containers/Data/Application/*/Library/Maps/GeoHistory.mapsdata','*/GeoHistory.mapsdata',),
+        "paths": (
+            '*/mobile/Containers/Data/Application/*/Library/Maps/GeoHistory.mapsdata',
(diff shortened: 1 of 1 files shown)


JamesHabben:
added a helper function to address the deep path testing to reduce the duplication of paths. this module could probably use a lot more work, but out of scope for this fix. thanks for finding and fixing these!

PR #1388, opened 2025-12-08
fix(netusage): handle exceptions during timestamp conversion

### LINKED ISSUE #1389: [Bug] netusage artifact parser crashes on invalid timestamp values
The `get_netusage` function in netusage.py encounters a critical error when processing network usage data containing invalid or NULL timestamps in the SQLite database. The parser crashes with a `ValueError`, preventing the extraction of any network usage data, even if valid records exist.

**Root Cause Analysis**
The issue stems from the `convert_ts_human_to_utc` function being called on raw data from the SQLite query without error handling.
1.  The SQLite query uses `datetime(...)` which may return empty strings or unexpected formats for NULL/0 values.
2.  These values are passed to `convert_ts_human_to_utc`.
3.  `datetime.strptime` raises a `ValueError` (e.g., `time data '' does not match format...`) or `TypeError`.
4.  Since there are no `try-except` blocks around these conversion calls, the exception propagates and terminates the artifact parsing immediately.

**Traceback**
```text
ValueError: time data '' does not match format '%Y-%m-%d %H:%M:%S'
Traceback (most recent call last):
  File "scripts/artifacts/netusage.py", line 56
    lastconnected = convert_utc_human_to_timezone(convert_ts_human_to_utc(row[0]),timezone_offset)
  File "scripts/ilapfuncs.py", line 1117, in convert_ts_human_to_utc
    dt = datetime.strptime(ts, '%Y-%m-%d %H:%M:%S')
```

**Affected File**
*   netusage.py (Lines 52-60 and 104-105 approx)

**Steps to Reproduce**
1.  Run iLEAPP against an iOS backup containing a `netusage.sqlite` database.
2.  Ensure the database contains rows where `ZTIMESTAMP` or `ZFIRSTTIMESTAMP` results in a NULL or invalid string when queried.
3.  Observe the crash during the "Network Usage" artifact parsing stage.

**Expected Behavior**
The parser should implement defensive programming: if a timestamp is invalid, it should be logged or set to a fallback value (e.g., "N/A"), allowing the parser to continue processing the remaining valid rows.

**Proposed Fix**
Wrap the timestamp conversion logic in `try-except` blocks to catch `ValueError` and `TypeError`, setting the value to `'N/A'` upon failure.

Tokenzrey:
This PR fixes a critical bug in netusage.py that causes the artifact parser to crash when processing `netusage.sqlite` databases containing invalid or NULL timestamps.

Previously, the parser would fail completely upon encountering a single invalid timestamp string (e.g., an empty string returned by SQLite's `datetime` function), resulting in total data loss for this artifact. This update implements defensive programming to handle these exceptions gracefully.

### Changes
*   **Added:** `try-except` blocks around `convert_ts_human_to_utc` calls in netusage.py.
*   **Logic:** If a `ValueError` or `TypeError` occurs during conversion, the timestamp is set to `'N/A'` instead of crashing the script.
*   **Result:** The parser now processes all valid rows and marks invalid timestamps clearly, ensuring maximum data extraction.

### Verification & Testing
I have created a comprehensive test suite (test_netusage_fix.py) to verify the fix and ensure no regressions.

*   **Test Suite:** 5/5 tests passed.
*   **Scenarios Covered:**
    *   Invalid Timestamp from SQLite (Empty strings/NULLs).
    *   Valid Timestamp Conversion (Regression test).
    *   None Value Handling.
    *   Full Row Processing (Simulation of the fixed loop).
    *   Comparison of Original (Crashes) vs. Fixed (Handles gracefully) code.

**Test Output Summary:**
```text
TEST SUMMARY
================================================================================
Invalid Timestamp from SQLite..................... [PASS]
Valid Timestamp Regression........................ [PASS]
None Value Handling............................... [PASS]
Full Row Processing (Fixed)....................... [PASS]
Original vs Fixed Code............................ [PASS]
================================================================================
Total: 5/5 tests passed
[SUCCESS] ALL TESTS PASSED - Fix is working correctly!
```

### Checklist
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have verified the fix with a local test script.
- [ ] The changes generate no new warnings.

### Related Issues
Closes #1389

### DIFF
--- a/scripts/artifacts/netusage.py
+++ b/scripts/artifacts/netusage.py
@@ -48,17 +48,26 @@ def get_netusage(files_found, report_folder, seeker, wrap_text, timezone_offset)
                 for row in all_rows:
                     if row[0] is None:
                         lastconnected = ''
-                    else: 
-                        lastconnected = convert_utc_human_to_timezone(convert_ts_human_to_utc(row[0]),timezone_offset)
-                    if row[1] is None:    
+                    else:
+                        try:
+                            lastconnected = convert_utc_human_to_timezone(convert_ts_human_to_utc(row[0]),timezone_offset)
+                        except (ValueError, TypeError):
+                            lastconnected = 'N/A'
+                    if row[1] is None:
                         firstused = ''
                     else:
-                        firstused = convert_utc_human_to_timezone(convert_ts_human_to_utc(row[1]),timezone_offset)
-                    if row[2] is None: 
+                        try:
+                            firstused = convert_utc_human_to_timezone(convert_ts_human_to_utc(row[1]),timezone_offset)
+                        except (ValueError, TypeError):
+                            firstused = 'N/A'
+                    if row[2] is None:
                         lastused = ''
                     else:
-                        lastused = convert_utc_human_to_timezone(convert_ts_human_to_utc(row[2]),timezone_offset)
-                    
+                        try:
+                            lastused = convert_utc_human_to_timezone(convert_ts_human_to_utc(row[2]),timezone_offset)
+                        except (ValueError, TypeError):
+                            lastused = 'N/A'
+
                     data_list.append((lastconnected,firstused,lastused,row[3],row[4],row[5],row[6],row[7],row[8],row[9],row[10],row[11]))
 
                 report.write_artifact_data_table(data_headers, data_list, file_found)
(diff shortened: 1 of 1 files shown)


JamesHabben:
Closing this as addressed by #1426, which rewrote the netusage artifact and fixed #1389 by avoiding SQLite datetime string conversion for these Core Data timestamps.

PR #1394, opened 2025-12-08
fix: Fix invalid date handling in network usage data processing

hamasfaa:
## Fix year out of range error in netusage artifact

Fixed `ValueError: year 0 is out of range` that occurred when processing timestamps with invalid or negative values in the netusage database.

**Before:** Code attempted to convert invalid timestamps (resulting in year 0 or negative years) without validation
**After:** Added validation to check for invalid timestamps (starting with '0000') and wrapped timestamp conversions in try-except blocks to catch ValueError and OSError. Invalid timestamps are now set to empty strings

### Error Messages Before Fix

ValueError: year 0 is out of range

**Files changed:**
- `scripts/artifacts/netusage.py`


JamesHabben:
closing as fixed in other PRs

PR #1396, opened 2025-12-08
Fix TikTok artifacts SQL parsing and add robust error handling

AdityaS207:
## Summary

This pull request fixes runtime failures when parsing TikTok artifacts by ensuring constructed SQL is syntactically valid and by adding robust exception handling around SQLite operations. The change prevents artifact crashes caused by missing contact tables or schema differences across TikTok database samples.

## Previously observed errors

- sqlite3.OperationalError: near ")": syntax error (caused by empty/invalid subquery e.g., LEFT JOIN ())

- sqlite3.OperationalError: no such table: TIMMessageKVORM

- These raised unhandled exceptions that stopped artifact processing and sometimes terminated the overall run.

### 1. tikTok Artifact

- Use ' UNION ALL '.join(...) when assembling contact subqueries so multiple AwemeContactsV% tables are combined correctly.

- Add a safe fallback subquery when no contact tables are found (a valid SELECT that returns zero rows) to prevent LEFT JOIN () syntax errors.

- Wrap main SQL executions in try/except sqlite3.OperationalError and log errors via logfunc(...) rather than letting exceptions crash the parser.

### 2. tikTokReplied Artifact

- Mirror the same robust subquery assembly and fallback behavior for contact tables.
- Add import sqlite3 and wrap execution of the CTE + SELECT in a try/except sqlite3.OperationalError block; on error the artifact logs the issue and continues with an empty result.
- Ensure the deduplication CTE receives a guaranteed-valid contacts_subquery (either the UNION ALL of tables or the zero-row fallback).

## General improvements

- Standardized, non-intrusive error handling for SQLite operations: Error messages now follow the pattern: Reading <artifact> had SQL error: <exception>

- All fixes preserve original parsing logic and output formats where possible; changes prioritize stability over failing the entire run.

### 1. Local verification: 
ran ileapp.py with the provided profile limiting to tikTok and tiktok_replied modules (example command used during testing was the same command the user runs with 13-4-1.tar).

### 2. Observed behavior:

- Prior crashes caused by malformed subqueries or missing tables no longer abort artifact processing.

- When contact tables exist, artifacts still return results as before.

- When contact tables are absent or queries fail due to schema differences, the artifacts log the SQL error and continue; the run finishes and report generation completes.

### 2. Artifacts are resilient to:

- missing AwemeContactsV% tables,

- missing TIMMessageKVORM table,

- SQL differences that would previously crash the module.

## Notes / Caveats

- The fallback subquery intentionally returns zero rows. That means contact-related columns will be empty when contact tables are genuinely absent; this is a deliberate tradeoff to maintain run continuity.

- If the SQLite build used in a DB does not support window functions (e.g., ROW_NUMBER() in the CTE), the try/except will catch the error and log it. If preferred, we can provide a compatibility fallback (group-by/aggregate-based deduplication) instead of relying on CTE/window functions.

- These changes are defensive and do not add new external dependencies.

### DIFF
--- a/myprofile.json
+++ b/myprofile.json
@@ -1,1 +1,0 @@
-{"leapp": "ileapp", "format_version": 1, "plugins": ["tikTok", "tiktok_replied"]}
--- a/myprofile.json
+++ b/myprofile.json
@@ -1,0 +1,1 @@
+{"leapp": "ileapp", "format_version": 1, "plugins": ["tikTok", "tiktok_replied"]}
--- a/scripts/artifacts/tikTok.py
+++ b/scripts/artifacts/tikTok.py
@@ -40,39 +40,44 @@ def get_tikTok(files_found, report_folder, seeker, wrap_text, timezone_offset):
             for table in contacts_tables:
                 contacts_subqueries.append(f'SELECT uid, customid, nickname, url1 FROM {table}')
 
-            contacts_subquery = '''
-            UNION ALL
-            '''.join(contacts_subqueries)
+            if contacts_subqueries:
+                contacts_subquery = ' UNION ALL '.join(contacts_subqueries)
+            else:
+                contacts_subquery = "SELECT NULL as uid, NULL as customid, NULL as nickname, NULL as url1 WHERE 0"
 
-            cursor.execute(f'''
-                select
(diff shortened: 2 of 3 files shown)


JamesHabben:
stale and conflicted. carried idea into new PR

PR #1400, opened 2025-12-08
Fix Multiple Artifacts Processing Errors

michaelgrace72:
### Summary
This PR fixes 5 errors that caused artifact processing to fail during iOS data extraction. All fixes have been tested successfully with iOS 15 Public Image

### Artifacts Fixed
1. chromeLoginData - Duplicate Column Error
The all_data_headers was adding 'Browser Name' even though it already exist in data_header.
Before:
`all_data_headers = lava_data_headers + ['Browser Name']`
After:
`all_data_headers = lava_data_headers`
2. cloudkitParticipants - Type Error
Code assumed participant['UserIdentity'] was always a dictionary, but it could be a string
Fix: Added type checking and safe dictionary access

3. cloudkitSharing - Type Error
Same issue as cloudkitParticipants - participant data structure varied.

4. FitnessWorkoutsLocationData - Missing Table
Code queried associations table without checking if it exists.
Fix: Added table existence check

5. health - Index Out of Range
Query returns varying column counts depending on database schema, but code assumed all indices exist.
Fix: Added bound checking

### DIFF
--- a/scripts/artifacts/FitnessWorkoutsLocationData.py
+++ b/scripts/artifacts/FitnessWorkoutsLocationData.py
@@ -28,7 +28,7 @@ def get_Health(files_found, report_folder, seeker, wrap_text, timezone_offset):
     
     cursor = db.cursor()
 
-    if  does_table_exist_in_db(healthdb_secure, 'location_series_data') == True:
+    if  does_table_exist_in_db(healthdb_secure, 'location_series_data') == True and does_table_exist_in_db(healthdb_secure, 'associations') == True:
     
     # Fitness Workouts Location Data Analysis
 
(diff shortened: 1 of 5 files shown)


JamesHabben:
there is a lot going on in this PR. it would be better submitted as PR for each module so the review can be focused. I think some of these issues have been resolved in other PRs and this is a lot of stuff to work through. please review the PR and see what is still relevant.

abrignoni:
Thanks! This conflicts with `main` now — `chrome.py` and several of these artifacts are mid-modernization to the LAVA pipeline. Could you rebase onto the latest `main` and confirm which fixes still apply? Some may already be handled by the newer versions. Happy to re-review after a rebase.

JamesHabben:
PR is in conflict and no response from submitter. items are all now addresses in separate PRs. closing this one.

PR #1402, opened 2025-12-09
Fix parsing errors in Alarms, Connected Devices, Plist, and NetUsage artifacts

alfathandr:
## Summary
This PR fixes several execution errors encountered when processing specific artifacts (Alarms, Connected Devices, System Version Plist, and Network Usage). These changes improve tool stability and prevent crashes due to data parsing issues.

### 1. Alarms Artifact (`appleAlarms.py`)
- **Issue:** Encountered `ValueError: Invalid isoformat string: '...Z'` because `datetime.fromisoformat()` does not support the 'Z' suffix (Zulu time) in certain Python versions.
- **Fix:** Switched to using the native `plistlib` date object and explicitly set the timezone to UTC using `.replace(tzinfo=timezone.utc)`.

### 2. Connected Device Information (`ConnectedDeviceInformation.py`)
- **Issue:** `sqlite3.DatabaseError: file is not a database` crashed the tool when processing corrupt or 0-byte SQLite files.
- **Fix:** Implemented a `try-except` block to catch `sqlite3.DatabaseError` and `sqlite3.OperationalError`. This allows the tool to log the error and skip the invalid file gracefully instead of crashing.

### 3. System Version Plist (`systemVersionPlist.py`)
- **Issue:** `TypeError` caused by variable shadowing. The variable name `system_version_plist` conflicted with a function/module name, causing Python to treat it as a function instead of a string path.
- **Fix:** Refactored the variable name to `plist_file` to ensure type safety.

### 4. Network Usage (`netusage.py`)
- **Issue:** `ValueError: year 0 is out of range`. The SQLite database contained timestamps with year 0, which is invalid in Python's Gregorian calendar implementation.
- **Fix:** Added a `try-except` block to handle conversion failures. If a timestamp is invalid, it falls back to the raw string value instead of crashing the artifact.

### DIFF
--- a/scripts/artifacts/ConnectedDeviceInformation.py
+++ b/scripts/artifacts/ConnectedDeviceInformation.py
@@ -105,11 +105,13 @@ def connected_device_info_device_history(context):
     return data_headers, data_list, source_path
 
 
+import sqlite3 # <--- PENTING: Tambahkan ini di bagian paling atas file jika belum ada
+
 @artifact_processor
-def connected_device_info_consolidated_connected_device_history(context):
+def connected_device_info_device_history(context):
     """
-    Extracts and consolidates device connection history from
-    'healthdb_secure.sqlite'.
+    Extracts and processes historical information about connected devices from
+    the `healthdb_secure.sqlite` database.
     """
 
     files_found = context.get_files_found()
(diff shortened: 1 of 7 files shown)


JamesHabben:
this is a very large PR and would be better done as a single module in each PR. additionally, other PRs have resolved some of the issues here. please review this PR and see what is still needed.

abrignoni:
Thanks for this! Since this was opened, `main` has moved a lot — several of these artifacts plus `ilapfuncs.py` changed during the ongoing LAVA/`@artifact_processor` modernization (e.g. `tikTok` was updated in #1545), and GitHub now flags this PR as conflicting. Could you rebase onto the latest `main` and confirm which fixes are still needed? Some may already be resolved. Happy to re-review once it's rebased. Thanks again!

abrignoni:
Thank you for working on these. mailprotect, netusage and systemVersionPlist were since fixed at the source, tikTok was rewritten, and core ilapfuncs changes are being held for the planned cores consolidation. The connected-devices addition would need its own PR against current main if you want to pursue it. Closing.

PR #1403, opened 2025-12-09
Fixing Handling Mail Protect

auldaf:
Add exception for mail protect


JamesHabben:
closing because this pr is messy with other modules and commits blended together, and the mail protect guard is better control in the central API already.

PR #1404, opened 2025-12-09
Fitness error handle

Vinoryz:
This pull request improves the robustness of the `get_Health` function in `FitnessWorkoutsLocationData.py` by enhancing the way required database tables are checked before analysis. The main change is that the function now checks for the existence of multiple required tables and logs any that are missing, instead of only checking for a single table.

Database validation improvements:

* The function now verifies that all required tables (`location_series_data`, `data_series`, `associations`, `workout_activities`) exist in the `healthdb_secure.sqlite` database before proceeding, and logs any missing tables.
* Removed the previous logic that only checked for the existence of the `location_series_data` table and logged a message if it was missing, making the validation process more comprehensive.


JamesHabben:
this was superseded by an earlier pr fix for the same issue in #1363

PR #1405, opened 2025-12-09
Fix: Handle missing columns in Health Source Devices

amadheatrizza:
This pull request improves the robustness of the `health_source_devices` artifact processor in `scripts/artifacts/health.py` by adding a check for required columns in the `source_devices` table before executing the main query. If any columns are missing, the function logs the missing columns and returns early, preventing possible runtime errors.

**Health source devices extraction improvements:**

* Added a check for required columns in the `source_devices` table using the new `does_column_exist_in_db` function; if any are missing, logs the missing columns and returns early to avoid runtime errors.
* Updated imports to include `does_column_exist_in_db` and `logfunc`, supporting the new column existence check and logging functionality.

### DIFF
--- a/scripts/artifacts/health.py
+++ b/scripts/artifacts/health.py
@@ -253,8 +253,8 @@ __artifacts_v2__ = {
 }
 
 from packaging import version
-from scripts.ilapfuncs import artifact_processor, get_sqlite_db_records, \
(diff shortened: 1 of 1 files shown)


JamesHabben:
Closing this as addressed by #1415 / commit 71a820a4. Current main already checks sync_provenance and sync_identity before adding them to the query and output. Treating those columns as required would skip valid older schemas, so the current optional-column approach is preferred.

PR #1408, opened 2025-12-09
Move data_headers scope

Vinoryz:
This pull request improves the `geodMapTiles` function in `geodMapTiles.py` by making its return values more consistent and robust, especially in cases where data is missing. It also ensures that headers are always defined and that users receive clearer feedback when no data is available.

Return value consistency and error handling:

* The function now consistently returns a tuple of `(data_headers, data_list, file_found)` even when an exception occurs due to missing columns, preventing downstream errors.
* When there are no rows to process, the function logs a message indicating that no data is available in the Map Tile Cache, improving user feedback.

Data structure and output improvements:

* The `data_headers` variable is now always defined at the start of data processing, ensuring headers are available for report generation and return values. 
* The function's return statement at the end now always returns `data_headers`, `data_list`, and `file_found`, aligning with its usage in other parts of the code.

### DIFF
--- a/scripts/artifacts/geodMapTiles.py
+++ b/scripts/artifacts/geodMapTiles.py
@@ -133,10 +133,13 @@ def geodMapTiles(files_found, report_folder, seeker, wrap_text, timezone_offset)
     except Exception as e:
         print(e)
         logfunc('Table is missing columns. No data available.')
-        return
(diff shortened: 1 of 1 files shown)


JamesHabben:
closing since there was no response and there are adjustments still needing to be done

PR #1409, opened 2025-12-09
Implement db checker and try-except block

Vinoryz:
This pull request improves the robustness and error handling of the `atxDatastore` artifact processor in `scripts/artifacts/atxDatastore.py`. The main focus is to ensure that missing or inaccessible databases are properly logged and handled, preventing unhandled exceptions during execution.

**Error handling and logging improvements:**

* Added checks to ensure both `ATXDataStore` and `Local` databases are present before proceeding, with clear log messages if either is missing.
* Added error handling for database opening and attachment failures, logging specific error messages and safely returning early if issues are encountered.
* Wrapped the SQL query execution in a try-except block to catch and log any exceptions during data retrieval, ensuring the database connection is closed properly on error.
* Added `logfunc` to the imports to enable logging functionality.

**Data processing adjustment:**

* Simplified timestamp conversion by removing unnecessary `None` checks, since the conversion function can handle `None` values.

### DIFF
--- a/scripts/artifacts/atxDatastore.py
+++ b/scripts/artifacts/atxDatastore.py
@@ -13,7 +13,7 @@ __artifacts_v2__ = {
     }
 }
 
-from scripts.ilapfuncs import artifact_processor, open_sqlite_db_readonly, attach_sqlite_db_readonly, convert_cocoa_core_data_ts_to_utc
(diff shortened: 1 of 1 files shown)


JamesHabben:
closing this. the error condition you found is valid, but this approach is not the right way. we have api functions that do all the error catching and logging.

fixed in #1539

PR #1410, opened 2025-12-09
Implement column checker

Vinoryz:
This pull request adds a check for required database columns in the `Ph6_1ViewandPlayDataPhDaPsql` artifact processor to ensure compatibility with newer iOS versions. The main change is to verify that the `ZLASTVIEWEDDATE` column exists in the `ZADDITIONALASSETATTRIBUTES` table before proceeding, which helps prevent errors when processing databases from iOS 16 and above.

Database compatibility improvements:

* Added a check using `does_column_exist_in_db` to ensure the `ZLASTVIEWEDDATE` column exists in the `ZADDITIONALASSETATTRIBUTES` table for iOS 16 and above, logging and exiting early if the column is missing.
* Imported the `does_column_exist_in_db` function from `scripts.ilapfuncs` to support the new column existence check.

### DIFF
--- a/scripts/artifacts/Ph6ViewedPlayData.py
+++ b/scripts/artifacts/Ph6ViewedPlayData.py
@@ -37,7 +37,7 @@ __artifacts_v2__ = {
 
 import os
 from packaging import version
-from scripts.ilapfuncs import artifact_processor, get_file_path, open_sqlite_db_readonly, get_sqlite_db_records, logfunc, iOS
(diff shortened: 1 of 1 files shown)


JamesHabben:
closing as this module has a more recent update PR by the author

abrignoni:
Follow-up: the column guard proposed here has landed on `main` via #1713 — `Ph006ViewedPlayData` (all three functions) now checks `ZADDITIONALASSETATTRIBUTES.ZLASTVIEWEDDATE` on iOS 16+ and skips cleanly on unmigrated databases instead of erroring. Thanks @Vinoryz for identifying the failure mode and the fix approach.

PR #1411, opened 2025-12-09
Implement column checker

Vinoryz:
This pull request improves the handling of database schema differences for iOS versions 16 and above in the `Ph51PossibleOptimizedAssetsIntResouPhDaPsql` artifact processor. The main change is the addition of a check to ensure required columns exist in the `ZADDITIONALASSETATTRIBUTES` table before proceeding, which helps prevent errors when processing databases with missing columns.

**Database schema validation for iOS 16+:**

* Added a check using `does_column_exist_in_db` to verify that the `ZDUPLICATEDETECTORPERCEPTUALPROCESSINGSTATE` column exists in the `ZADDITIONALASSETATTRIBUTES` table for iOS versions 16 and above, logging a message and exiting early if the column is missing.
* Imported the `does_column_exist_in_db` utility from `scripts.ilapfuncs` to support the new column existence check.

### DIFF
--- a/scripts/artifacts/Ph51PossOptimizedAssetsIntResouData.py
+++ b/scripts/artifacts/Ph51PossOptimizedAssetsIntResouData.py
@@ -22,7 +22,7 @@ __artifacts_v2__ = {
 
 import os
 from packaging import version
-from scripts.ilapfuncs import artifact_processor, get_file_path, open_sqlite_db_readonly, get_sqlite_db_records, logfunc, iOS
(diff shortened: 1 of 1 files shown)


JamesHabben:
closing as this module has a more recent PR update from the author

abrignoni:
Follow-up: the column guard proposed here has landed on `main` via #1713 — `Ph051PossOptimizedAssetsIntResouData` now checks `ZADDITIONALASSETATTRIBUTES.ZDUPLICATEDETECTORPERCEPTUALPROCESSINGSTATE` on iOS 16+ and skips cleanly on unmigrated databases instead of erroring. Thanks @Vinoryz for identifying the failure mode and the fix approach.

PR #1412, opened 2025-12-09
Fix AttributeError crash in user_defaults artifact parsing

prnndk:
When processing artifacts, the user_defaults module crashed with an `AttributeError: 'str' object has no attribute 'items'.`

This occurs because, in some instances, the plist loader returns a string (e.g., when a previous exception handler sets the plist content to `"INVALID FILE"`) or fails to load a dictionary structure, but the code attempts to iterate over it using `.items()` without validation.

## Changes Made
Added Type Validation: Implemented `isinstance(plist, dict)` checks before attempting to iterate through `plist.items()`. This ensures that only valid dictionary objects are processed.

Enhanced Error Handling: Wrapped the file processing loop in a try-except block to gracefully handle InvalidFileException or other generic exceptions.

Safety Checks: Added a check for MCMMetadataIdentifier to ensure the plist contains the necessary metadata before processing, preventing potential KeyErrors.

Test Result: 
image

### DIFF
--- a/scripts/artifacts/userDefaults.py
+++ b/scripts/artifacts/userDefaults.py
@@ -31,15 +31,22 @@ def user_defaults(files_found, report_folder, seeker, wrap_text, timezone_offset
         file_found = str(file_found)
         if not file_found.endswith('mobile_container_manager.metadata.plist'):
             continue
-        with open(file_found, "rb") as fp:
-            if sys.version_info >= (3, 9):
-                plist = plistlib.load(fp)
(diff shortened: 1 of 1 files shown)


JamesHabben:
Thanks for the fix. This crash has already been addressed on current main by validating plist content is a dictionary before iterating it, and the module has since been modernized to use the shared plist helper and v2 context API. Closing this PR because it is stale and would revert newer module changes.

PR #1413, opened 2025-12-09
fix: skip invalid user defaults plists

Ricardo08S:
Summary
- Harden Application User Defaults artifact against invalid plist files so corrupt entries are logged and skipped instead of crashing the entire artifact.

Implementation Details
- Imports logfunc and emits a clear message when a plist fails to parse (both plistlib and biplist code paths).
- Ensures only dictionaries are iterated, preventing AttributeError when the file is malformed or contains unexpected types.
- Keeps the rest of the reporting pipeline unchanged.

### DIFF
--- a/scripts/artifacts/userDefaults.py
+++ b/scripts/artifacts/userDefaults.py
@@ -20,7 +20,7 @@ import plistlib
 from plistlib import InvalidFileException
 import sys
 import datetime
-from scripts.ilapfuncs import artifact_processor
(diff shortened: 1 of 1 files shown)


JamesHabben:
Thanks for the fix. Current main already skips invalid/non-dictionary plist content in userDefaults using the shared plist helper, so this crash is addressed. Closing because this PR is stale and would revert the module from the newer v2 context/shared-helper implementation.

PR #1428, opened 2026-01-29
DroneFlightDAT

nopiadintya:
DroneFlightDAT-.DAT file parser

### DIFF
--- a/scripts/artifacts/droneFlightLog.py
+++ b/scripts/artifacts/droneFlightLog.py
@@ -20,13 +20,11 @@ import datetime
 from scripts.ilapfuncs import artifact_processor, logfunc
 
 @artifact_processor
-def droneFlightLog(files_found, report_folder, seeker, wrap_text, timezone_offset):
+def droneFlightLog(files_found, _report_folder, _seeker, _wrap_text, _timezone_offset):
     data_list = []
     source_paths = set()
-
-    sorted_files = sorted(files_found, key=lambda x: os.path.getmtime(x))
+    sorted_files = sorted(files_found, key=os.path.getmtime)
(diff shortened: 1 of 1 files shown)


JamesHabben:
@nopiadintya if you can share some sample file listings we can help build a better search pattern

abrignoni:
Thanks for the DJI drone log parser! Two things to fix before this can be merged:

1. Crash on no-data: `droneFlightLog` does `return None` when no entries are parsed, but the `@artifact_processor` wrapper unpacks the return value (`data_headers, data_list, source_path = func(...)`), so returning `None` raises `TypeError: cannot unpack non-iterable NoneType`. Please return an empty result instead, e.g. `return data_headers, [], ''`.

2. Path glob too broad: `paths: ("**/*.log",)` matches every `.log` file anywhere in the extraction, so the artifact opens and reads all of them on every case. Please scope it to the DJI drone app's container/path (bundle id or known DJI log directory) so it only runs on relevant files.

Once those are addressed I'll re-review. Thanks!

nopiadintya:
Hello sir, I sincerely apologize for the delayed response. I haven't been
able to reply sooner due to some health issues, and I've also been
experiencing technical issues with the device I'm using for this project. I
will address your feedback on the code as soon as I can. Thank you for your
patience, your understanding, and for reviewing the parser

abrignoni:
Thank you for taking a swing at drone logs. This one cannot merge as is: the search pattern matches every .log file on a device with nothing anchoring it to DJI, the timestamps combine each line's time with the file modification date, which on an extracted archive is the extraction date, and the parser handles text logs while the title promises DAT. Closing, and contributions are always welcome.

PR #1463, opened 2026-04-01
Update trustedpeers module to include more useful information

ghmihkel:
Add device colors and extract peer serial numbers from ZESCROWMETADATA.ZPEERINFO.


JamesHabben:
would you be willing to incorporate some other upgrades as part of this PR? we have a guide here: https://github.com/abrignoni/iLEAPP/blob/main/admin/docs/module_updates.md

JamesHabben:
no response and the code needs lots of adjustment. good find on new information though, so I took that stuff and made a new PR as #1541

PR #1469, opened 2026-04-23
instagram-threads-update

Sukochev:
Added __artifacts_v2__ 
Removed __artifacts__
Removed ArtifactHtmlReport and updated function to remove report generation (followed module_updates.md instructions)
Removed unused imports

Ran "black" against file which resulted in formatting updates (functional changes are presented below).

Updated userPk variable path (received error in this path during testing -- was the inspiration for update).

Updated userFull variable path.
Updated inviterPk & inviterFull paths (interestingly Instagram renamed the key to "IGUser*inviter_DEPRECATED".)

Added shared_media_url_expiration_date to output. 

Module worked in my local testing. 
All the best!
Regards
Greg

### DIFF
--- a/scripts/artifacts/instagramThreads.py
+++ b/scripts/artifacts/instagramThreads.py
@@ -14,6 +14,22 @@ __artifacts_v2__ = {
         ),
         "output_types": ["html", "tsv", "timeline"],
         "artifact_icon": "instagram",
+    },
+    "get_instagram_calls": {
+        "name": "Instagram Threads Calls",
+        "description": "Existing calls sent and received in the Instagram App.",
+        "author": "Alexis Brignoni",
+        "version": "0.7",
+        "creation_date": "2021-03-09",
+        "last_update_date": "2026-04-23",
+        "requirements": "",
+        "category": "Instagram",
+        "notes": "",
+        "paths": (
+            "*/mobile/Containers/Data/Application/*/Library/Application Support/DirectSQLiteDatabase/*.db*"
+        ),
+        "output_types": ["html", "tsv", "timeline"],
+        "artifact_icon": "phone",
     }
 }
 
@@ -48,7 +64,6 @@ def get_instagramThreads(
             usageentries = len(all_rows)
             userdict = {}
             data_list = []
-            video_calls = []
 
             if usageentries > 0:
                 for row in all_rows:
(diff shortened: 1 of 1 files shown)


Sukochev:
Working on it

Sukochev:
Put code under for loop to deal with undefined loop variable.
Added _ to unused arguments. Removing arguments caused errors in ileapp.py.
Removed fila=0, does not appear to affect results. 
Added _ to unused variables. 

Module worked in local testing. 
Regards,
Greg

Sukochev:
More testing required. Closed for now.

Sukochev:
Returned backward compatibility. Returned Calls functionality.
Tested against JH iOS 17 image to make sure results returned were the same as the previous script.

JamesHabben:
check out the document here on the things we target when updating a module to the current core framework. 
https://github.com/abrignoni/iLEAPP/blob/main/admin/docs/module_updates.md
- it looks like some headers can use a `datetime` designation 
- function parameters can be updated to use `context`
- change the output to `standard` so it includes lava output.

also, I note that the `return` statement is inside the loop, so this will collect data from the first db file found and then exit. it will not process any further db files and it will not make it to the `db.close()` line either. is this intended or is it possible to have multiple db files on a device if there are multiple accounts in the app?

also, there are a couple videos toward the bottom of the page that may be of help. one shows how the messages artifact can be built to build the conversation view in lava. if the conversation stuff is overwhelming, dont worry about it. we can adjust later.
https://www.leapps.org/resources

Sukochev:
Thanks James. I have updated the script. Added datetime designation and tested output in both html and lava app. All looks ok in html still, however getting datetime issues in lava app view. The datetime values in the database/plist itself come back as UTC datetime.datetime objects (e.g. `datetime.datetime(2026, 5, 13, 4, 31, 13)`) after `get_plist_content()` function (`nska_deserialize`) is run. However, when looking at the report in LAVA app, the app seems to adjust for my local timezone automatically (it changes the timestamp to be 10.hrs earlier even though the timestamp is originally/already UTC). Although I have not tested the `convert_plist_date_to_utc` function against the timestamp values yet.

JamesHabben:
sounds like an issue with not declaring the timestamp as UTC, so the date functions assume you are giving a local timezone value. so either specifically declare the timezone as UTC, or preferably use the `convert_plist_date_to_utc` function that we made to hopefully be easier for developers to use in modules.

Sukochev:
`convert_plist_date_to_utc` worked nicely to fix the UTC timestamp discrepancy, Thanks! 
I added conversation view in LAVA and it appears to have worked (at least in the two extractions I tested).

JamesHabben:
thanks for putting in all that work and testing it!

PR #1484, opened 2026-05-19
feature: add command line option to point to additional artifact modu…

JamesHabben:
…le folder

- adjustments to core leapp and plugin_loader code to accommodate for an additional artifact folder to load modules from
- the primary artifact folder is still used, modules in additional folder are added to the list of artifacts
- any module in alternate location with duplicate key will override the module in the standard folder
  - this allows for testing with a modified module while using a built release version of the tool
- added `image_test.py` as a test-only artifact to generate testing outputs
  - artifact intentionally uses LEAPP media manager to create outputs
- sample image types added to `/admin/test/samples_data/` with a manifest for the test artifact parsing


JamesHabben:
this now allows for the benefit of testing image parsing and detection. in the sqlite db, we can see all the image types that are from the test, and that svg has an issue with the mime type.
image

JamesHabben:
this is what this test artifact creates for lava
image

JamesHabben:
@Johann-PLW I think you said you were good with this PR?

JamesHabben:
spoke with @Johann-PLW and he said this is good to merge.

PR #1488, opened 2026-05-24
Notes: Drop imghdr from Notes artifact

### LINKED ISSUE #1246: Make "standard-imghdr" a requirement for Python 3.13 compatibility
I observed that iLEAPP works without any problems in a Python 3.13 environment when the “standard-imghdr” package is installed via pip. (Testing on Manjaro Linux x86-64)

This could be added as an optional dependency for Python 3.13 and above like this:

`standard-imghdr; python_version >= "3.13"`

Further testing is required for macOS and Windows environments.

### LINKED ISSUE #1321: standard-imghdr missing
Running under Linux Ubuntu, i am getting this error:
...
File "/media/shared/data/iLEAPP/scripts/artifacts/notes.py", line 7, in <module>
    import imghdr
ModuleNotFoundError: No module named 'imghdr'

After I added _standard-imghdr_ in requirements.txt, everything is fine

BR, Jens

Mirochill:
Closes #1246.
Closes #1321.

Summary:
- Replace the Notes artifact's deprecated `imghdr.what()` checks with the existing `scripts.filetype.image_match` helper.
- Keep thumbnail creation limited to JPEG/PNG attachments while avoiding the removed stdlib `imghdr` module.

Validation:
- `git diff --check`
- `rg -n "\bimghdr\b|imghdr\.what" . --glob "!*.ipynb"`
- Not run locally: tests/build/lint.

### DIFF
--- a/scripts/artifacts/notes.py
+++ b/scripts/artifacts/notes.py
@@ -4,11 +4,11 @@
 
 from os.path import dirname, join
 from PIL import Image
-import imghdr
 import zlib
 import binascii
 
 from scripts.artifact_report import ArtifactHtmlReport
+from scripts.filetype import image_match
 from scripts.ilapfuncs import logfunc, tsv, timeline, open_sqlite_db_readonly, does_column_exist_in_db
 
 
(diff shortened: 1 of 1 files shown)


JamesHabben:
this is marked as a draft still, but it looks like a valid issue to address. @Mirochill did you want to update and remove conflicts?

abrignoni:
Thanks @Mirochill — the direction here was right, and this helped surface the issue. It's since been overtaken by the LAVA conversion of `notes.py` (commit f79def50), which removed `import imghdr` entirely and replaced the thumbnail logic with `check_in_embedded_media(...)`. That embeds every attachment type rather than only jpg/png, so it's a superset of the `image_match` approach proposed here, and `imghdr` no longer appears anywhere in the module — there's nothing left to drop.

Closing as superseded, with thanks. The underlying requirement (#1246) and the crash it caused (#1321) are both addressed on `main`.

PR #1495, opened 2026-05-27
Fix bug in batteryBDC

JamesHabben:
- the watts column was causing an index error on csv files where there were not enough columns
- check for 9 columns in source file; write error and skip file if less
- safely get watts value or '' if column isn't there
- lint error cleanup

### DIFF
--- a/scripts/artifacts/batteryBDC.py
+++ b/scripts/artifacts/batteryBDC.py
@@ -5,7 +5,7 @@ __artifacts_v2__ = {
         "description": "Parses battery usage and temps from Battery Data Collection (BDC) logs",
         "author": "@stark4n6",
         "creation_date": "2026-03-18",
-        "last_update_date": "2026-03-18",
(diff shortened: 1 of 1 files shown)


JamesHabben:
here is the error i got when running against josh's ios15 image. for reference

```bash
[21/439] batteryBDC [batteryBDC] artifact started
Reading batteryBDC artifact had errors!
Error was list index out of range
  File "ileapp.py", line 506, in crunch_artifacts
  File "scripts\ilapfuncs.py", line 477, in wrapper
    data_headers, data_list, source_path = func(Context)
                                           ^^^^^^^^^^^^^
  File "C:\Users\James Habben\AppData\Local\Temp\_MEI247162\scripts\artifacts\batteryBDC.py", line 47, in batteryBDC
    watts = item[18]
            ~~~~^^^^
IndexError: list index out of range
```

PR #1503, opened 2026-05-29
discordChat: fix bug that caused multiplied duplicate records

JamesHabben:
also linter cleanup

### DIFF
--- a/scripts/artifacts/discordChats.py
+++ b/scripts/artifacts/discordChats.py
@@ -1,252 +1,382 @@
 __artifacts_v2__ = {
     "discordChats": {
         "name": "Discord - Chats",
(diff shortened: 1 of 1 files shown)


stark4n6:
@JamesHabben looks like one more linter thing to clean up, I guess "attachments" variable isn't used

JamesHabben:
@stark4n6 I did notice that one but didn't have the time or data sources to fully dig in. The bug I fixed here was taking 60 records and turning it into 62,000 because of the logic in the loop. I wanted to get that taken care of. 

I also noticed that there is a lot of duplication/overlap between the discordChats and discord_a module, so I was planning to come back around on that later.

JamesHabben:
@stark4n6  this pretty much makes discord_a duplicative / obsolete, but I didn't delete it yet

PR #1506, opened 2026-05-29
fix bug in integration test harness that errored on new module

JamesHabben:
the test harness was designed to grab git commit info to include with the various files that it creates to keep track of testing versions. these test scripts assumed that the module being tested against had been committed to git already, but i run into an error because i was trying to test a local module that i had not yet added to git. this checks for that and gracefully moves forward.


JamesHabben:
I added a bunch of doc strings and fixed *some* of the linter errors. its ready now.

PR #1509, opened 2026-05-31
update ileappGUI to better handle monitor space on macOS

JamesHabben:
- center case data modal window to the main ileappGUI window (all OS)
- detect screen and open ileappGUI on detected screen where multiple monitors exist (macOS)
  - previously it was opening in unexpected positions with part of window off screen (macOS)
- use `grab_set_global()` to better capture focus on case data modal (macOS)
  - tk docs say global is not safe to use on windows or linux
- add macOS height for case data window (macOS)

### DIFF
--- a/ileappGUI.py
+++ b/ileappGUI.py
@@ -414,24 +414,47 @@ def case_data():
 
     ### Case Data Window creation
     case_window = tk.Toplevel(main_window)
+    case_window.transient(main_window)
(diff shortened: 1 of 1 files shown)


JamesHabben:
I have tested this on Mac and windows. have not tested on linux, but not much code should impact running there.

also, specifically avoided linter error fixes to keep this diff clean.

PR #1511, opened 2026-06-01
Zalo Chat and Users Parser

prosch88:
This adds a parser for the app Zalo (most commonly used in Vietnam or by people with ties to Vietnam).

parses:
- Chats
- Groupchats
- known Users

image

Some newer versions of the app use JPEG XL. To view them, you'll need a modern browser (currently mostly still in beta) or the latest version of LAVA.

### DIFF
--- a/scripts/artifacts/ZaloChats.py
+++ b/scripts/artifacts/ZaloChats.py
@@ -98,8 +98,9 @@ def build_gif(png_files, metadata_file):
     return output
 
 @artifact_processor
-def zalo_users(files_found, _report_folder, _seeker, _wrap_text, _timezone_offset):
+def zalo_users(context):
     """Extracts known Zalo Users"""
+    files_found = context.get_files_found()
     data_list = []
     source_path = get_file_path(files_found, 'profile.sqlite')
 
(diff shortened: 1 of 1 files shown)


prosch88:
Thanks to @BrunoFischerGermany for the test data sets!

JamesHabben:
2 small comments
1. if you use the newer `context` processor preamble in the functions you dont have to underscore the unused parameters. then use `context.get_files_found()` to replace the `files_found` parameter of the older processor function preamble.
2. I see a `mobile` column header. is that possible a `phonenumber` type of data?

prosch88:
You're right. It is indeed a phone number. I didn't notice the switch to "context." It's definitely easier to read.

JamesHabben:
The context parameter also makes the modules more resilient to core code updates since changing the preamble of the processor function requires updates to every module. We can update context with new features now and not trigger a massive module update.

prosch88:
I had changed it. It still works with my test data. So it should be fine, right?

JamesHabben:
yes. take another look through and it looks good to me so merged it in base on you saying your tests still work. thanks for contributing!

PR #1512, opened 2026-06-01
Refactor platform metadata lookups in Context for cross LEAPP

JamesHabben:
This refactor moves platform-specific reference data out of shared Python logic and into explicitly named JSON collections. RLEAPP may need to resolve both Apple and Android device metadata from account-return records, while iLEAPP and ALEAPP primarily use their own platform data.

Keeping the Context metadata API generic makes the shared framework code easier to synchronize directly between LEAPP repositories. Each project can include only the JSON collections it needs, while reusable collections can be copied between projects without adding repository-specific branching.

Where not possible (the apple version mess), platform specific function code can live across platforms with a graceful failure if the data file needed for the code is missing.

- Replace Apple-specific JSON loaders with generic cached metadata helpers: `get_metadata()` and `lookup_metadata()`.
- Rename metadata files to describe their mapping direction: `apple_device_id_to_model.json`, `apple_board_id_to_model.json`, and `apple_build_id_to_os_version.json`.
- Add `get_apple_os_version()` for Apple device-family routing and the existing modern iPadOS label adjustment.
- Update affected artifacts to use the new generic lookups and Apple OS resolver.
- Add metadata documentation


JamesHabben:
An example from RLEAP where there is a similar list of apple devices as what we have here, but this list is not as comprehensive nor is it even up to date. 
https://github.com/abrignoni/RLEAPP/blob/e84a66ac3057579f15dda6928202f1926176a8f0/scripts/artifacts/takeoutLocationHistorySettings.py#L54

With this PR, RLEAPP can be build to use the same `apple_device_id_to_model.json` file and keeping RLEAPP up to date is as simple as a file copy (and can even be automated).

Johann-PLW (approved):
Everything's fine on my side.
Perfect to use it across the LEAPPs.

stark4n6:
@JamesHabben @Johann-PLW looks like there are conflicts with os_builds.json but I can't see them properly

JamesHabben:
probably from the recent add @Johann-PLW made to the builds file data. I will take a look. the web conflict resolver is saying its too complex so will have to pull it locally and mess with it.

PR #1523, opened 2026-06-09
Parse sms's `attributedBody` using `pytypedstream`

davidyuk:
# Summary
This PR addresses an issue where many SMS and iMessage texts are missing or extracted as empty strings in the final reports. It introduces `attributedBody` extraction using the `pytypedstream` library as a fallback when the standard `message.text` field is empty.

# Context & Motivation
In modern iOS versions (especially iOS 15, 16, and newer), Apple increasingly defaults to storing message content within the `message.attributedBody` field instead of the classic `message.text` column. This behavior is triggered when messages involve:
- Advanced formatting or screen effects.
- Mentions, inline links, or complex emojis.
- Edited or unsent messages (introduced in iOS 16).
- Interactions with iMessage apps/extensions.

Currently, iLEAPP only extracts data from `message.text`. As a forensic consequence, reports miss a significant portion of actual conversation logs, rendering those records blank.
By integrating `pytypedstream`, we can safely unarchive Apple's proprietary `typedstream` (`NSMutableAttributedString`) binary blobs, extract the pure textual data across all global encodings, and restore the missing chat context without breaking backward compatibility.

# Testing Done
- Tested against recent iOS backups containing mixed-language chats (including non-ASCII scripts and emojis).
- Verified that messages encoded into `attributedBody`, which previously resulted in blank rows, are now fully populated in both the HTML rendering and the output data lists.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 astc_decomp_faster
+pytypedstream
 bencoding
 biplist
 blackboxprotobuf
--- a/scripts/artifacts/sms.py
+++ b/scripts/artifacts/sms.py
@@ -28,13 +28,12 @@ __artifacts_v2__ = {
     }
 }
 
-import os
 import pandas as pd
 
 from scripts.artifact_report import ArtifactHtmlReport
 from scripts.chat_rendering import render_chat, chat_HTML
 from scripts.ilapfuncs import artifact_processor, get_file_path, get_sqlite_db_records, \
-    convert_cocoa_core_data_ts_to_utc, check_in_media, logfunc, lava_get_full_media_info
+    convert_cocoa_core_data_ts_to_utc, check_in_media, lava_get_full_media_info
 
 import typedstream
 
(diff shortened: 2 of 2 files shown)


JamesHabben:
without having inspected this yet, I am wondering if there is value to parsing this always. is there a chance that data in the stream is meaningfully different than the text column?

abrignoni:
@JamesHabben it parses the area if there is nothing in message.text.

JamesHabben:
Yes I understand that. I am wondering if there is any scenario where the text column has data but the stream has more or different data. In the current logic, we would only grab the text.

davidyuk:
I understand the concern. This is a partial solution. There can be some enhanced message types, but I'm not using iMessage to check that. But it works well for usual SMS (including line wraps, links, emojis) stored in `attributedBody`.

JamesHabben:
that makes sense. enhanced would be great in the future, but not critical now since it sounds like the core of the message would be the same if its in the text column.

PR #1531, opened 2026-06-12
update LinkedIn module to conversation

JamesHabben:


stark4n6:
@JamesHabben shouldn't the get source files be just using the full file name? Also it doesn't appear this is taking -wal into account in the regex

PR #1536, opened 2026-06-17
modernize geodMapTiles and convert heic tiles to jpg

JamesHabben:


JamesHabben:
tested against josh ios17 image which has 12 heic tile images to extract and view

PR #1538, opened 2026-06-18
Small fix for app snapshots

maala-nfi:
Since `_lava_media_info.updated_at` is the second to last column in the result of the query used in function `lava_get_full_media_info`, the list item used for variable `last_modified_date` should be updated from -1 to -2. (The query was updated in commit 28ea00e9cbe3d8e87192adbe3dda84a08fef603f.)


JamesHabben:
Good catch on `updated_at` moving positions.

Could we avoid the positional index here and use the row column name instead?

maala-nfi:
> Could we avoid the positional index here and use the row column name instead?

Yes! That would be even better. 
I tested it and committed the change.

PR #1543, opened 2026-06-18
Add keychain artifact module

kobo220:
Artifact module that parses keychain data from an iTunes backup or the two major vendor acquisition formats. There are five different reports generated:

- Wifi credentials
- Web passwords
- Device Bluetooth information
- Paired Bluetooth device information
- Mail accounts

Code tested against various public test images, CTF images, and personal test images from iOS 13-26.

### DIFF
--- a/scripts/artifacts/keychain.py
+++ b/scripts/artifacts/keychain.py
@@ -1,0 +1,901 @@
+__artifacts_v2__ = {
+    "keychain_wifi": {
+        "name": "Keychain Wifi Credentials",
+        "description": "Parses keychain to extract stored Wi-Fi credentials",
+        "author": "@kobo220",
+        "creation_date": "2026-06-18",
+        "last_update_date": "2026-06-18",
+        "requirements": "none",
+        "category": "Keychain",
+        "notes": "",
+        "paths": (
+            '*/keychain-backup.plist',
+            '*/extra/KeychainDump/backup_keychain_v2.plist',
+            '*/Keychains/keychain-2.db*',
+        ),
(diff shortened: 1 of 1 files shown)


JamesHabben:
making a note here that @snoop168 has some related stuff in #1280 but looks like some different approaches. looking to get thoughts from y'all on how these can move forward together.

also, what public images are out there with data to test against this keychain stuff? I dont have access to any of the keychain extraction commercial tools right now to make my own.

kobo220:
For public images I'm using Josh Hickman's images and images found on CFReDS (primarily Magnet and Cellebrite CTF images):
https://cfreds.nist.gov/
https://digitalcorpora.s3.amazonaws.com/s3_browser.html#corpora/mobile/ 

I'm going to look over the code in that other PR and do some testing. I do prefer the idea of processing/storing the keychain data in a more core location so it can be used more readily in other artifacts. I have another artifact I'm almost done with for Signal, that was making use of my keychain data by sloppily importing `from scripts.artifacts.keychain import parse_keychain`.

JamesHabben:
Thanks for the image info. 

Importing code from another module should be alright, but importing data would be risky as we can't guarantee order of module execution. Just a note.

snoop168:
I think my code was made to the core ileapp portion so that the keychain can be used by the modules for parsing. Session depended on having the key in the keychain. I'm sure there are many other artifacts that would benefit from that code in the core. I haven't looked at this code yet but even though it's probably touching and parsing the same file I think it has a different goal.

Issue #1496, opened 2026-05-27
Error in biomeAppinstall


JamesHabben:
running against MVS2026 iPhone14 image creates this error log:

```bash
[34/439] get_biomeAppinstall [biomeAppinstall] artifact started
Reading get_biomeAppinstall artifact had errors!
Error was 7
Exception Traceback: Traceback (most recent call last):
  File "ileapp.py", line 506, in crunch_artifacts
  File "scripts\ilapfuncs.py", line 479, in wrapper
    data_headers, data_list, source_path = func(files_found, report_folder, seeker, wrap_text, timezone_offset)
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\James Habben\AppData\Local\Temp\_MEI152962\scripts\artifacts\biomeAppinstall.py", line 49, in get_biomeAppinstall
    protostuff, types = blackboxprotobuf.decode_message(record.data, typess)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "blackboxprotobuf\lib\interface.py", line 62, in decode_message
  File "blackboxprotobuf\lib\types\length_delim.py", line 228, in decode_message
  File "blackboxprotobuf\lib\types\length_delim.py", line 294, in decode_lendelim_message
  File "blackboxprotobuf\lib\types\length_delim.py", line 245, in decode_message
  File "blackboxprotobuf\lib\types\length_delim.py", line 309, in decode_group
  File "blackboxprotobuf\lib\types\length_delim.py", line 166, in decode_message
KeyError: 7
```

@snoop168 i think this is yours?

Gear-I:
Currently troubleshooting the crash within the parser

Gear-I:
Fix biomeAppinstall protobuf decode crash

PR #1546, opened 2026-06-18
Fix biomeAppinstall protobuf decode crash

Gear-I:
Fixes biomeAppinstall.py parser file crashing when blackboxprotobuf encounters unsupported wire type 7. The parser now skips only the affected record and continues processing valid records. 
This parser has been tested on iPhone 14 Pro FFS test image and Belkasoft | iOS 16.3 test iOS image. 
This fix is related to issue #1496


Gear-I:
Created a new branch for BiomeAppInstall 
Created a new PR (Fix biomeappinstall keyerror7
#1547)

PR #1548, opened 2026-06-18
Fix biomeappinstall keyerror7

Gear-I:
Fixes biomeAppinstall.py parser file crashing when blackboxprotobuf encounters unsupported wire type 7. The parser now skips only the affected record and continues processing valid records.
This parser has been tested on iPhone 14 Pro FFS test image and Belkasoft | iOS 16.3 test iOS image.
This fix is related to issue https://github.com/abrignoni/iLEAPP/issues/1496


scripts/artifacts/biomeAppinstall.py
```diff
@@ -19,6 +19,7 @@
 from scripts.ccl_segb.ccl_segb import read_segb_file
 from scripts.ccl_segb.ccl_segb_common import EntryState
 from scripts.ilapfuncs import artifact_processor, webkit_timestampsconv
+from scripts.ilapfuncs import artifact_processor, logfunc
```
JamesHabben: this is creating a duplicate import. add logfunc to the line above instead.

scripts/artifacts/biomeAppinstall.py
```diff
@@ -90,52 +91,98 @@ def get_biomeAppinstall(context):
...
+
         for record in read_segb_file(file_found):
             ts = record.timestamp1
             ts = ts.replace(tzinfo=timezone.utc)
 
             if record.state == EntryState.Written:
+                try:
+                    protostuff, _ = blackboxprotobuf.decode_message(record.data, typess)
+                except (KeyError, ValueError, TypeError, IndexError) as ex:
+                    logfunc(f"Skipping biomeAppinstall record due to protobuf decode error: {ex}")
```
JamesHabben: we should include additional information about the context of the error. 
- `context.get_relative_path(file_found)` should give file path info
- `record.data_start_offset` should give further insight into offset within that file

Gear-I: Should I include it in the actual file as a comment?

scripts/artifacts/biomeAppinstall.py
Gear-I: image

Added file path and record offset to the exception logging as requested. Re-tested against my iOS test image. The artifact completes successfully and reports the specific records causing the protobuf decode failure.

JamesHabben:
closes #1496

PR #1621, opened 2026-06-26
facebookMessenger-update

Sukochev:
This update was inspired by a recent acquisition where another database contained data of interest. Added artifact Facebook Messenger - Client Messages. Also sought to update script for other artifacts to use context, however it appears that was done at the same time (two days ago), so apologies for the reformatting. 
Testing on JH iOS17 yielded a pleasant surprise that "Secret Conversations" were picked up, so new artifact may have legacy value.
Testing other artifacts confirmed output was not altered. 

Regards
Greg

### DIFF
--- a/scripts/artifacts/facebookMessenger.py
+++ b/scripts/artifacts/facebookMessenger.py
@@ -1,108 +1,180 @@
 __artifacts_v2__ = {
-    'facebookMessengerCalls': {
-        'name': 'Facebook Messenger - Calls',
-        'description': 'Extract call history from Facebook Messenger.',
-        'author': '@stark4n6',
-        'creation_date': '2021-03-03',
-        'last_update_date': '2025-04-09',
-        'requirements': 'none',
-        'category': 'Facebook Messenger',
-        'notes': '',
-        'paths': (
-            '*/lightspeed-userDatabases/*.db*',),
-        'output_types': 'standard',
-        'artifact_icon': 'phone-call',
+    "facebookMessengerCalls": {
+        "name": "Facebook Messenger - Calls",
+        "description": "Extract call history from Facebook Messenger.",
+        "author": "@stark4n6",
+        "creation_date": "2021-03-03",
+        "last_update_date": "2025-04-09",
+        "requirements": "none",
+        "category": "Facebook Messenger",
+        "notes": "",
+        "paths": ("*/lightspeed-userDatabases/*.db*",),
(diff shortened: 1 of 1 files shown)


abrignoni:
Thanks Greg — really nice addition, especially the new **Client Messages** artifact (and the bonus Secret Conversations recovery on iOS 17). Before merging I ran a review and have a few correctness items I'd like to see addressed, since they're the kind of thing a single-device test wouldn't surface.

**1. Multi-database regression (most important).** The new helpers return on the *first* matching DB:

```python
def _database_files_with_view(context, view):
    for file_found in context.get_files_found():
        if str(file_found).endswith(("db")) and does_view_exist_in_db(file_found, view):
            return [str(file_found)]   # returns on first hit
```

The previous version accumulated **all** matching `.db` files into `db_path_list` and ran `get_sqlite_multiple_db_records`. With this change, an acquisition containing more than one `lightspeed-userDatabases` DB (e.g. multiple Messenger accounts on a device) will silently process only one and drop the rest. This affects Calls, Chats, Secret Conversations, Conversation Groups and Contacts — not just the new artifact. Could you restore the all-DB behavior (accumulate every match, then loop over them)?

**2. `COALESCE` dropped in the Chats direction logic.** Current `main` has:

```sql
COALESCE(contacts.name, '') || ' (Local User)'
```

This PR changes it to `contacts.name || ' (Local User)'`. In SQLite `NULL || '...'` evaluates to `NULL`, so when the local user's contact name is null the sender name comes back empty instead of "(Local User)". The same pattern was copied into the new client-chats query. Worth restoring the `COALESCE`.

**3. `endswith(("db"))` / `endswith(("jpg"))` aren't tuples.** The extra parens are a no-op — these are just `endswith("db")` / `endswith("jpg")`, which match any path ending in the literal "db"/"jpg". The original used `.endswith('.db')`. Low risk, but worth tightening to `'.db'` / `'.jpg'`.

Everything else looks good — it's read-only (no evidence writes), and the media handling correctly goes through `check_in_media`. Once the multi-DB aggregation and the `COALESCE` are back, this is good to go.

---
*Note for transparency: this review/analysis was produced by Claude (Claude Code) at my request. — Brigs*

Sukochev:
Applied fixes and tested with more than one database in path location. 
Regards
Greg

abrignoni:
Thank you for this artifact. Super appreciated. 🙏🏽 -Brigs

PR #1630, opened 2026-06-27
feat: new module for home depot app artifacts

JamesHabben:


JamesHabben:
ew, why are we failing on TODO hits

abrignoni:
> ew, why are we failing on TODO hits

Because the linter is stupid. LOL!!!

PR #1631, opened 2026-06-28
feat: allow custom output folder name

JamesHabben:
- note: CLI and framework functionality was already in place
- GUI now has textbox to allow customization of output folder name
- guards illegal folder name characters in both CLI and GUI
- guards to prevent overwriting existing folder in CLI and GUI
- illegal character file/folder functions moved from `ilapfuncs` to new modularized `/leapp_functions/` path
- `ilapfuncs` imports old functions from new path to keep backwards compatibility
- left commented out def lines in `ilapfuncs` to note they have been moved

### DIFF
--- a/ileapp.py
+++ b/ileapp.py
@@ -18,6 +18,7 @@ from shutil import copy2
 from getpass import getpass
 from scripts.search_files import *
 from scripts.ilapfuncs import *
+from leapp_functions.output import validate_output_folder_available
 from scripts.version_info import leapp_name, leapp_version
 from time import process_time, gmtime, strftime, perf_counter
 from scripts.lavafuncs import *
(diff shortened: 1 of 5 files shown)


JamesHabben:
image

stark4n6:
@JamesHabben do we want the label to say something like "Custom Folder Name (optional)"?

Johann-PLW:
> @JamesHabben do we want the label to say something like "Custom Folder Name (optional)"?

I'm not sure it is relevant to rename the label as it is the name of the folder where the reports will be saved: Output_path + folder_name.

By default, it uses the convention we've always used: leappAppName_Output_YYYY-MM-DD_NameOfTheDay_HHMMSS.
By the way, I think we could remove the name of the day.

JamesHabben:
I agree with removing the day name. Makes it extra long for sure

PR #1632, opened 2026-07-01
Add Biome NotifcationUsage artifact

thysvanmeurs:
Hello!

This script parses the Notification Usage biome. It is a first version, I expect more to come as I will continue to validate the fields further. I did not want to sit on the script for too long, but rather share it with the rest of the community.

### DIFF
--- a/scripts/artifacts/biomeNotificationUsage.py
+++ b/scripts/artifacts/biomeNotificationUsage.py
@@ -8,7 +8,7 @@ __artifacts_v2__ = {
         "requirements": "none",
         "category": "Biome",
         "notes": "",
-        "paths": ('*/biome/streams/restricted/Notification.Usage/*'), 
+        "paths": ('*/Biome/streams/restricted/Notification.Usage/*'),
         "output_types": "standard"
     }
 }
@@ -19,13 +19,13 @@ from datetime import timezone
 import blackboxprotobuf
 from scripts.ccl_segb.ccl_segb import read_segb_file
 from scripts.ccl_segb.ccl_segb_common import EntryState
-from scripts.ilapfuncs import artifact_processor, webkit_timestampsconv, convert_utc_human_to_timezone
+from scripts.ilapfuncs import artifact_processor, webkit_timestampsconv, logfunc
 
 
 @artifact_processor
-def get_biomeNotificationUsage(files_found, report_folder, seeker, wrap_text, timezone_offset):
+def get_biomeNotificationUsage(context):
 
-    typess = { 
+    typess = {
                 '1': {'name': '', 'type': 'str'}, '2': {'name': '', 'type': 'double'}, '3': {'name': '', 'type': 'int'},
                 '4': {'name': '', 'type': 'str'}, '5': {'name': '', 'type': 'str'}, '8': {'name': '', 'type': 'str'}, '9': {'name': '', 'type': 'str'},
                 '11': {'name': '', 'type': 'int'}, '12': {'name': '', 'type': 'str'}, '14': {'name': '', 'type': 'str'},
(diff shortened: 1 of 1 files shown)


scripts/artifacts/biomeNotificationUsage.py
```diff
@@ -0,0 +1,83 @@
...
+                    else:
+                        notificationContent2 = None
+                        
+
+                    data_list.append((ts, timeStart, record.state.name, appName, notificationContent, notificationContent2, filename, record.data_start_offset))
+        except ValueError as e:
+            print(f"Skipping {file_found}: {e}")
+            continue
+    data_headers = ('SEGB Timestamp', 'datetime', 'SEGB State', 'Application name', 'Notification content', 'Notification content 2', 'Filename', 'Offset')
+    return data_headers, data_list, report_file
```
abrignoni, line 69: Bundled all the fixes here so it applies atomically in one commit (rationale in the review summary). Tested locally — compiles, registers as **Biome - Notification Usage**, Context-form, 8 columns with both timestamps datetime-typed.

```suggestion
__artifacts_v2__ = {
    "get_biomeNotificationUsage": {
        "name": "Biome - Notification Usage",
        "description": "Parses the Notification Usage biome",
        "author": "Thijs van Meurs",
        "version": "0.0.1",
        "date": "2026-06-29",
        "requirements": "none",
        "category": "Biome",
        "notes": "",
        "paths": ('*/Biome/streams/restricted/Notification.Usage/*'),
        "output_types": "standard"
    }
}


import os
from datetime import timezone
import blackboxprotobuf
from scripts.ccl_segb.ccl_segb import read_segb_file
from scripts.ccl_segb.ccl_segb_common import EntryState
from scripts.ilapfuncs import artifact_processor, webkit_timestampsconv, logfunc


@artifact_processor
def get_biomeNotificationUsage(context):

    typess = {
                '1': {'name': '', 'type': 'str'}, '2': {'name': '', 'type': 'double'}, '3': {'name': '', 'type': 'int'},
                '4': {'name': '', 'type': 'str'}, '5': {'name': '', 'type': 'str'}, '8': {'name': '', 'type': 'str'}, '9': {'name': '', 'type': 'str'},
                '11': {'name': '', 'type': 'int'}, '12': {'name': '', 'type': 'str'}, '14': {'name': '', 'type': 'str'},
                '16': {'name': '', 'type': 'int'}, '18': {'name': '', 'type': 'int'}
              }

    data_list = []
    report_file = 'Unknown'
    for file_found in context.get_files_found():
        file_found = str(file_found)
        filename = os.path.basename(file_found)
        if filename.startswith('.'):
            continue
        if not os.path.isfile(file_found) or 'tombstone' in file_found:
            continue
        report_file = os.path.dirname(file_found)

        try:
            for record in read_segb_file(file_found):
                ts = record.timestamp1.replace(tzinfo=timezone.utc)

                if record.state == EntryState.Written:
                    protostuff, _ = blackboxprotobuf.decode_message(record.data, typess)

                    raw_ts = protostuff.get('2')
                    timeStart = webkit_timestampsconv(raw_ts) if raw_ts is not None else None
                    appName = protostuff.get('14')
                    notificationContent = protostuff.get('8')
                    notificationContent2 = protostuff.get('12')

                    data_list.append((ts, timeStart, record.state.name, appName,
                                      notificationContent, notificationContent2, filename,
                                      record.data_start_offset))
        except ValueError as e:
            logfunc(f"Skipping {file_found}: {e}")
            continue

    data_headers = (('SEGB Timestamp', 'datetime'), ('Timestamp', 'datetime'), 'SEGB State',
                    'Application name', 'Notification content', 'Notification content 2',
                    'Filename', 'Offset')
    return data_headers, data_list, report_file
```

abrignoni (changes requested):
Thanks for this, Thijs — nice work on the SEGB/protobuf decode, and thanks for sharing the first pass rather than sitting on it. To line it up with the current Biome artifacts I've bundled the changes into a single commit-able suggestion below (one click replaces the file).

**Main changes**
- **Context form** (`def get_biomeNotificationUsage(context):`) — the other 22 Biome artifacts all use it.
- **Store timestamps as UTC.** The Context form has no `timezone_offset`, `webkit_timestampsconv` already returns tz-aware UTC, and the LAVA viewer applies the selected zone — so `convert_utc_human_to_timezone` is dropped (otherwise `Timestamp` ends up in local time while `SEGB Timestamp` stays UTC, and it double-offsets in LAVA once the column is typed).
- **`data_headers`**: both timestamps as nested `('name', 'datetime')` tuples. As written the flat `'datetime'` became its own column and neither timestamp was datetime-typed.
- **Path casing** `biome` → `Biome`. Matching is case-sensitive (`os.path.normcase` is a no-op off Windows) and the real path is `…/Library/Biome/streams/…`, so the lowercase glob wouldn't hit a normal extraction.
- **Guard the protobuf reads** with `.get()` — a `KeyError` on `protostuff['2']`/`['14']` isn't caught by the `except ValueError` and would crash the artifact.
- `print(...)` → `logfunc(...)`.

I tested the result locally: it compiles, registers as **Biome - Notification Usage**, and resolves as the Context form. One thing worth a quick confirm against a real image is the `Biome` casing — I went with capital to match the documented path and the 21 sibling artifacts. Thanks again for contributing! 🙏

thysvanmeurs:
Dear mr. Brignoni,

Thank you for your quick reply and thorough feedback! I have approved the changes and it is ready to merge.

abrignoni:
Thank you, Thijs! 🙏 Really appreciate you sharing this with the community early instead of sitting on it — a brand new Biome stream (`Notification.Usage`) that nothing else parses yet. Merging it in.

A couple of friendly notes for whenever you pick this back up (no rush, totally fine as a first version):

- **Deleted records:** right now the parser only keeps `EntryState.Written` entries. The sibling `biomeNotificationsPub.py` also handles the `EntryState.Deleted` branch (appending the row with null fields), which recovers deleted notification usage — often the most interesting stuff for us forensically. Adding that would also give your "SEGB State" column something to show beyond "Written." 🙂
- **Metadata keys:** most of the Biome family now uses `creation_date`/`last_update_date` (and an `artifact_icon`) rather than `version`/`date` — worth aligning on the next pass for consistency.

Neither is a blocker. Excellent contribution and looking forward to the field validation you mentioned. Thanks again! 👏

PR #1634, opened 2026-07-01
Add: Oura Ring artifact parser

Slay3r00:


abrignoni:
Really nice contribution, @slay3r00 — thorough coverage across account/identity, Find My Ring GPS, ring/battery status, activity timeline, heart rate, and the Segment analytics, with solid UTC handling and nested bplist/JSON/NSDecimal decoding. 👏

I converted the 7 functions to the modern **context form** (`def x(context)` + `context.get_files_found()`) in place of the legacy 5-arg signature, and marked the defensive broad excepts with an inline pylint disable — that takes it to **10.00/10** and green CI with no behavior change. Merging.

A few **non-blocking** follow-ups for whenever you revisit it:

1. **`ts_auto` date-ish heuristic (worth a look — forensic weight).** A non-timestamp integer whose key matches `date|time|_at|seen|…` and falls in the 5e8–1e13 range would be rendered as a timestamped event in the activity timeline. The range/key guards keep incidence low, but since this is a forensic tool, a *mislabeled* timestamp is the failure mode most worth guarding against — worth confirming against real extractions and/or tightening.
2. **Silent excepts (lines ~235 and ~304).** These swallow parse failures with no trace; a `logfunc(...)` like the one at line 248 would make failures visible when debugging odd extractions.
3. **`ts_iso` fallback.** When it can't parse, it returns the raw string, which can land in a `datetime`-typed column. LAVA tolerates it, so lowest priority — just noting it.

None of these block the merge. Thanks again for the solid parser!

---
*For transparency: the context-form cleanup and this review were done with Claude (Claude Code) at my request. — Brigs*

JamesHabben:
I will add some other notes that are hand-typed. :P

you are using module level date conversion functions where we have coverage with api functions in place. we prefer modules do this to have more consistent date handling.
- `ts_cocoa` should use `convert_cocoa_core_data_ts_to_utc`
- `ts_unix` should use `convert_unix_ts_to_utc` or `convert_unix_ts_in_seconds`
- `ts_iso` against plist dates should use `convert_plist_date_to_utc`
- `ts_iso` against iso strings doesn't have a converter right now

you are also doing manual plist content loading and we have a couple plist api functions that will safely load and decode a lot of the structure including NSKeyed decoding.
`get_plist_file_content` for plist files
`get_plist_content` for plist data from blobs

nice work on the research and code so far!

Slay3r00:
Thank you both for the feedback! I planned on making a couple more for eventbrite app and dropbox app, and will use the recommended methods provided.

PR #1635, opened 2026-07-01
chore: update readme content

JamesHabben:
- clean up old stuff
- remove the module dev examples
- direct viewers to releases first
- document command arguments
- links to leapps website and other resources


abrignoni (approved):
This is good. Lots of old stuff needed to be removed. Perfect.

PR #1499, opened 2026-05-28
iOS26 Photos sqlite Support and Parser Description Updates

ScottKjr3347:
-Updated most parsers to support iOS26
-Updated Parser Names and Description to help with what parser can pull -Removed old parsers from artifacts folder
-Added more parsers to exclude list to speed up processing

PR #1641, opened 2026-07-02
HTML report: match runtime report names to artifact icons

abrignoni:
## Problem
Chromium artifacts (and the threaded SMS report) show the **alert-triangle** icon in the HTML report sidebar, and the same Chromium entries fall back to the default icon in LAVA.

Two distinct causes, same root: these reports are written under **runtime names** with custom output calls, not the artifact metadata name:
1. **HTML**: the sidebar resolves icons by joining `(category, report filename)` against icons registered under the metadata name — `chrome.py` writes `Chrome - Web History` / `Edge - Cookies` / … while the icon is registered under `Web History` / `Cookies`; `sms.py` writes `SMS & iMessage - Messages (Threaded)` vs `SMS`. Lookup miss → `alert-triangle`.
2. **LAVA**: `chrome.py`'s 17 per-browser `lava_process_artifact(...)` calls omitted the `artifact_icon` kwarg, so no icon was ever stored in the LAVA database and the viewer used its neutral default.

## Fix
- `scripts/report.py`: on an exact-match miss, fall back to the **longest** registered artifact name the report filename starts or ends with. Exact-name lookups unchanged; unknown names still fall through to `alert-triangle`.
- `scripts/artifacts/chrome.py`: pass `artifact_icon=__artifacts_v2__[module_name].get("artifact_icon")` at all 17 per-browser LAVA call sites, matching what the `@artifact_processor` wrapper does.

## Verified
- Simulated against the real loader-registered icons: all per-browser Chromium names resolve (`Chrome - Web History → history`, `Opera - Cookies → cookie`, `Browser - Login Data → key`, `Edge - Network Action Predictor → trending-up`, …); `SMS & iMessage - Messages (Threaded) → message`; exact matches (`Web History`, `Biome - Notifications`, `Calendar Events`, …) unchanged
- All 17 chrome.py sites confirmed passing the icon; `chromeWebHistory` resolves to `history` at import
- `py_compile` clean; `pylint --disable=C,R` on both files: **10.00/10**

## Known leftovers (intentional)
- `Ph10`/`Ph70` custom report names (`Ph10.1-Assets have embedded files-PhDaPsql`) don't relate to their metadata names and keep the triangle — both files are removed by #1499, so this resolves itself.

### DIFF
--- a/scripts/artifacts/chrome.py
+++ b/scripts/artifacts/chrome.py
@@ -359,7 +359,8 @@ def chromeWebHistory(files_found, report_folder, _seeker, wrap_text, _timezone_o
             category = "Chromium"
             module_name = "chromeWebHistory"
             table_name, object_columns, column_map = lava_process_artifact(category, module_name, report_name,
-                                                                           lava_data_headers, len(data_list))
+                                                                           lava_data_headers, len(data_list),
+                                                                           artifact_icon=__artifacts_v2__[module_name].get("artifact_icon"))
 
             lava_insert_sqlite_data(table_name, data_list, object_columns, lava_data_headers, column_map)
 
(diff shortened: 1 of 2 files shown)


abrignoni:
@Johann-PLW @JamesHabben for your awareness when porting to rest of LEAPPs.

JamesHabben:
`context.get_artifact_info()` would get you the exact artifact info. @abrignoni 
image

JamesHabben:
this module could use a lot more overhaul, especially to context parameter.

also found a bug where `chromeOfflinePages` is claiming `"chromeTopSites"` as its module name, so it will overwrite the actual top sites artifact. moving this dynamic naming over to using the context parameter and its getters will resolve copy/paste bugs like this.

PR #1643, opened 2026-07-02
Modernize burnerCache: context form, remove dead global chain, fix latent crash

abrignoni:
## Summary

Modernizes `burnerCache.py` (original artifact by @djangofaiola — parsing logic unchanged) and removes the last `global` statement in the iLEAPP artifact tree.

### Context form
All four artifact functions converted from the legacy 5-arg signature to the modern context form (`def x(context)` + `context.get_files_found()` / `context.get_seeker()`).

### Global removal — turned out to be dead code
Tracing `burner_app_identifier` end to end:
- Written only by `burnerCache_preferences` (an artifact with `output_types: ["none"]` and an empty return — its *sole* job was setting the global).
- Read only by one `seeker.search()` in `burnerCache_messages`… whose result (`media_files`) **was never used anywhere**. The fsCachedData files it searched for are already collected via the artifact's own `paths` glob and resolved later through `get_file_path`.

So the global, the search, and the `burnerCache_preferences` artifact were all a dead chain — deleted rather than refactored. burnerCache artifact count goes 5 → 4 (the removed one produced no report, TSV, LAVA, or log output, so nothing user-visible changes).

### Latent crash fixed
`burnerCache_numbers`: when a number has a `settings` dict but no `autoReplyMessage`, `auto_reply_enabled`/`auto_reply_text` were returned without ever being assigned → `NameError` (pylint E0606). Now default to `'No'`/`None` before the conditional.

### Lint 9.04 → 10.00/10
Unused locals removed (`artifact_info`, `artifact_info_name` ×4, the never-read image/thumbnail placeholder block), `filter` builtin no longer shadowed, f-strings without interpolation fixed, the two defensive broad excepts inline-disabled, dead imports dropped.

### Left in place (documented follow-up)
`burner_uid_map` — the module-level dict that accounts/contacts/numbers populate and numbers/messages read for ID→number resolution. It's the same cross-artifact shared-state pattern, but readers degrade gracefully (fall back to raw IDs), and fully de-globalizing it means re-parsing Cache.db per artifact — a behavior-sensitive refactor I didn't want to make **without Burner test data to validate output parity**. Added a comment documenting the ordering dependency instead.

## Testing
- `PYTHONPATH=. pylint scripts/artifacts/burnerCache.py --disable=C,R` → **10.00/10** (from 9.04).
- Byte-compiles; PluginLoader registers all 4 artifacts in context form (597 plugins total, no errors).
- CRLF line endings preserved (file is CRLF on main) — diff is the real ~111 lines, not a whole-file rewrite.
- No other module imports from burnerCache (grep-verified).

### DIFF
--- a/scripts/artifacts/burnerCache.py
+++ b/scripts/artifacts/burnerCache.py
@@ -1,19 +1,4 @@
 __artifacts_v2__ = {
-    "burnerCache_preferences": {
-        "name": "Burner Cache Preferences",
-        "description": "Parses and extract Burner Cache Preferences",
-        "author": "@djangofaiola",
-        "version": "0.3",
-        "creation_date": "2024-03-05",
-        "last_update_date": "2025-05-02",
-        "requirements": "none",
-        "category": "Burner Cache",
-        "notes": ("https://djangofaiola.blogspot.com",
-                  "App version tested: 4.0.18, 4.3.3, 5.3.8, 5.4.11"),
-        "paths": ('*/mobile/Containers/Data/Application/*/Library/Preferences/com.adhoclabs.burner.plist'),
-        "output_types": [ "none" ],
-        "artifact_icon": "settings"
-    },
     "burnerCache_accounts": {
         "name": "Burner Cache Accounts",
         "description": "Parses and extract Burner Cache Accounts",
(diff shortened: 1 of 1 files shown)


JamesHabben:
Convert chat to conversation @abrignoni

PR #1646, opened 2026-07-03
Fix packaged builds: bundle leapp_functions & assets, resolve report assets when frozen

abrignoni:
## Problem

Building the CLI from `scripts/pyinstaller/ileapp_macOS.spec` on current `main` produces a **broken binary**. Two independent bugs, plus a size issue:

### 1. `leapp_functions` is not bundled → plugin loader crashes on startup
`scripts/artifacts/apple_atx_images.py` does:
```python
from leapp_functions.parsers.apple_atx import decode_atx_file
```
Artifacts are loaded **dynamically** (`plugin_loader.py`), so PyInstaller's static analysis never sees `leapp_functions` and doesn't bundle it. Running the packaged binary:
```
ModuleNotFoundError: No module named 'leapp_functions.parsers'
[ileapp] Failed to execute script 'ileapp' due to unhandled exception!
```
This aborts the whole run (the loader imports every artifact at startup).

### 2. Report assets are resolved from `Path.cwd()` → no icons in a packaged binary
`scripts/report.py` loads its Tabler icons and correction JSON with paths relative to the current working directory:
```python
Path.cwd().joinpath("assets/tabler_icons")
get_json_file_content('scripts/data/tabler_icon_correction.json')
```
That only works when iLEAPP is launched from the repo root. A packaged binary run from anywhere else logs:
```
Error: JSON file 'scripts/data/tabler_icon_correction.json' not found.
Error: File '.../assets/tabler_icons/home.svg' not found.
```
and generates a report with missing icons. The three CLI specs also never bundled `assets/` at all (the GUI specs already do).

### 3. Size — unused heavy libs bundled
`scipy`, `matplotlib`, and `pyarrow` are imported by **zero** iLEAPP files (they come in transitively via `pandas`). They add a lot to the binary for nothing.

## Fix

- **`scripts/report.py`** — resolve `assets/` and `scripts/data/` from `sys._MEIPASS` when frozen, falling back to `cwd` from source (no behavior change when run from source). New `_base_dir()` helper.
- **All six specs** (`ileapp[_macOS/_Linux].spec`, `ileappGUI[_macOS/_Linux].spec`) — add `leapp_functions` to `datas` + a hidden import for `leapp_functions.parsers.apple_atx`, and put the repo root on `pathex` so analysis can find it.
- **CLI specs** — also bundle `assets/` (the GUI specs already did).
- **All six specs** — add `scipy`, `matplotlib`, `pyarrow` to `excludes`.

## Testing

Built the macOS CLI (`ileapp_macOS.spec`) and verified end-to-end:
- Loads all 295 artifacts, including `apple_atx_images`, with **no import error**.
- Full parse of a test extraction **from an unrelated working directory**: report generates with Tabler icons rendered and **no missing-asset errors**.

The other five specs received the analogous change but were not built in this environment (no Windows/Linux here). They share the identical gaps, so the same fix applies.

### DIFF
--- a/scripts/pyinstaller/ileapp.spec
+++ b/scripts/pyinstaller/ileapp.spec
@@ -3,10 +3,11 @@
 block_cipher = None
 
 a = Analysis(['..\\..\\ileapp.py'],
-             pathex=['..\\scripts\\artifacts'],
+             pathex=['..\\scripts\\artifacts', '..\\..'],
              binaries=[],
-             datas=[('..\\', '.\\scripts')],
+             datas=[('..\\', '.\\scripts'), ('..\\..\\assets', '.\\assets'), ('..\\..\\leapp_functions', '.\\leapp_functions')],
              hiddenimports=[
+        'leapp_functions.parsers.apple_atx',
                 'astc_decomp_faster',
                 'bencoding',
                 'blackboxprotobuf',
@@ -25,7 +26,7 @@ a = Analysis(['..\\..\\ileapp.py'],
                 ],
              hookspath=['./'],
              runtime_hooks=[],
-             excludes=[],
+             excludes=['scipy', 'matplotlib', 'pyarrow'],
              win_no_prefer_redirects=False,
              win_private_assemblies=False,
              cipher=block_cipher,
--- a/scripts/pyinstaller/ileappGUI.spec
+++ b/scripts/pyinstaller/ileappGUI.spec
@@ -3,10 +3,11 @@
 block_cipher = None
 
 a = Analysis(['..\\..\\ileappGUI.py'],
-             pathex=['..\\scripts\\artifacts'],
+             pathex=['..\\scripts\\artifacts', '..\\..'],
              binaries=[],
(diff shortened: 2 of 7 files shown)


abrignoni:
Closing as superseded.

The spec-file half of this landed on `main` in 4d0c4fca ("fix: declare the imports frozen builds crash without"), which bundles `leapp_functions` and `assets` in all six specs, additionally declares `google.protobuf` and `PIL`, and adds a regression test in `admin/test/scripts/test_pyinstaller_unifiedlog_bundle.py`.

The `scripts/report.py` half is obsolete: icon loading was rewritten to use the Tabler webfont CSS (`scripts/_elements/tabler-icons.css`), `assets/tabler_icons/` was removed from the repo, and the remaining data paths resolve via `Path(__file__).resolve().parent`, which is already frozen-safe. The lint failure on this PR points at a line that no longer exists on `main`.

The only piece not applied is `excludes=['scipy', 'matplotlib', 'pyarrow']`. Those three are still unimported on `main`, but none is in `requirements.txt` or pulled transitively, so the exclusion is defensive only. If it's worth having, it belongs in a small PR against current `main` rather than a rebase of this branch across seven conflicting files.

PR #1673, opened 2026-07-07
Change ilapfuncs.get_sqlite_db_records() to a generator

bconstanzo:
This PR implements the changes discussed in issues #1433 and #1306, and goes beyond what PR #1318 proposed.

I tried to keep changes minimal everywhere, there are a few NOTE comments where I noticed things that could be improved, left for a sunnier day in the future.

Not sure it's in the final and best form possible, but it is a step in the general direction to work with generators.


JamesHabben:
I added the leveling tag since this will need to get applied to the other leapps. remove that tag once done.

PR #1678, opened 2026-07-08
fix: media checkin skip file open if forced mime

JamesHabben:
@Johann-PLW this is not a priority for the release. its an edge case where the media file checkin was opening a file even when the force mime and extension was being provided. wasted ops.

### DIFF
--- a/scripts/ilapfuncs.py
+++ b/scripts/ilapfuncs.py
@@ -245,6 +245,8 @@ def _check_in_media(media_id, source_path, is_embedded, name, media_data=None, c
             suffix = name.split('.')[-1]
         elif not is_embedded and len(source_path.split('.')[-1]) < 5:
             suffix = source_path.split('.')[-1]
+        elif media_data is None:
+            return None
         else:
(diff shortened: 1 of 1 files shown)


abrignoni:
Thanks, @JamesHabben. The wasted read was real, and it ended up fixed from the other side in #2220, which caps that read at the 8,192 bytes the type sniffer actually looks at instead of reading the whole file. Closing this on that basis.

Measured on an iOS 17 full filesystem image, with a profile of only the artifacts that pass `force_type`, so the best case for this change: 2,056 `check_in_media` calls, the forced-type condition here would skip the read on all 2,056 and the already-checked-in condition on none of them, and those reads now cost 79 ms out of a 48 second run. The same 1,661 files were 236.6 MB of reads before #2220 and are 16.6 MB after.

Against that saving, this adds a media reference lookup on every call, and `_check_in_media` runs the same two lookups again right after.

If you want the last of it, the narrower version is the forced-type skip on its own, with the extension test living in one place instead of being spelled out in both functions. Happy to look at that if you open it.

PR #1712, opened 2026-07-13
Fix case-sensitive path mismatch in Biome artifacts (biome -> Biome)

mb4n6:
Six Biome artifacts glob for */biome/streams/..., but the real iOS path is Biome (uppercase). With case-sensitive fnmatch, these never match and silently log No file found. (17 other Biome artifacts already use Biome.)

Fixed paths in: biomeInfocus, biomeDKInfocus, biomeSafari, biomeBattperc, biomeDevplugin, biomeLocationactivity.

Tested: App.InFocus stream → before No file found, after 5 records parsed.


abrignoni:
thank you for catching this.

PR #1722, opened 2026-07-17
Update united_airlines.py

stark4n6:
- Update regex to be more broad for different types of extractions
- Update last update date


JamesHabben:
Looks good to me

PR #1729, opened 2026-07-19
Add Apple Intelligence Report parser

malwr4n6:
## Description

This PR adds a new artifact plugin to parse **Apple Intelligence Report** JSON files exported from iOS, iPadOS and macOS devices.

Apple Intelligence Reports can be exported via:
- **iOS/iPadOS:** Settings → Privacy & Security → Apple Intelligence Report
- **macOS:** System Settings → Privacy & Security → Apple Intelligence Report

Available on iOS 18.4+ / macOS 15.4+.

## What it parses

The plugin extracts two types of records:

**Model Requests** (on-device inference):
- Timestamp (UTC + Local)
- Event ID
- User Trigger (Summarize, Compose, Key Points, List, Table, Knowledge, Visual Intelligence, Image Generation, Memory Creation, Assistant)
- Use Case (Raw)
- Source App (Siri, Writing Tools, Image Playground, Photos, Safari, Mail, Messages, Notes)
- Request (prompt text, extracted and cleaned from template/JSON formats)
- Response
- Model

**Private Cloud Compute (PCC) Requests**:
- Timestamp (UTC + Local)
- Request ID
- Pipeline Kind
- Adapter / Model
- Node count and validation status

## Platform detection

Automatically detects whether the report was exported from **iOS/iPadOS** or **macOS** based on the presence of Visual Intelligence triggers (camera-based, iOS/iPadOS only).

## Artifact paths

```
*/Apple_Intelligence_Report*.json
```

## Testing

Tested against Apple Intelligence Report JSON files exported from:
- macOS 15.4+
- iOS 18.4+

## Author

Bhargav Rathod ([@malwr4n6](https://github.com/malwr4n6)) — https://malwr4n6.com


malwr4n6:
## Testing Note

While testing locally on macOS with the compiled `ileappGUI.app` (v2026.1.0), I encountered the following error when attempting to use `-t file` input type via CLI:

```
Error on argument -o (input type)
```

**Environment:**
- macOS (Apple Silicon)
- iLEAPP v2026.1.0 (compiled `.app` binary)
- Input: single `Apple_Intelligence_Report.json` file

This appears to be a limitation of the compiled binary rather than the plugin itself. Running from the Python source (`ileapp.py`) via the iLEAPP source repo resolves the issue.

Happy to provide additional testing details or adjust the plugin if needed.

abrignoni:
Thank you for this one, very timely artifact. I pushed a fixup to your branch before merging: the metadata now uses the field names the framework reads, the Local Timestamp column had to go (the framework passes a timezone name, not minutes, so that column always fell back to raw epochs; the report layer handles rendering from the UTC datetime), and the prompt/response truncation is removed since the full text is the evidence. Notes now say the examiner supplies the exported file and none of our test images carry one, so the parser rests on your reports plus a synthetic round-trip which passes. Merging. Good stuff.

PR #1734, opened 2026-07-20
Oura: anchor parsing to the Oura container and report source files

### LINKED ISSUE #1727: Oura Ring Parser paths too generic
@Slay3r00 
The Oura ring parsers are parsing oura ring artifacts in my test data but I do not have Oura ring. Some of the paths are too generic and are including data from other applications. See below screenshot. It appears to me that these paths may be the issue as they are similarly named for other applications (including life360 which you see in my screenshot):

*/Containers/Data/Application/*/Library/Preferences/com.segment.storage.*.plist'
*/Containers/Data/Application/*/Library/Preferences/com.zendesk.*.plist'
*/Containers/Data/Application/*/Library/Preferences/????????-????-????-????-????????????.plist'

Image

Slay3r00:
Fixes #1727

The com.segment.storage.*, com.zendesk.* and account-UID plist search
patterns also match same-named SDK plists inside other apps' containers,
so the parsers emitted foreign app data (Life360) on extractions with no
Oura app. Parsing is now anchored to the directory that holds the
Oura-only com.ouraring.oura.plist and nothing is parsed without it; the
deviceUid fallback that let foreign files through is removed.

Also surface data provenance in the reports: per-row Source File column
on the merged identity artifact (extraction-relative), all contributing
plists listed in its source_path, and the analytics artifacts now report
the segment plist they actually read instead of the suite plist.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/scripts/artifacts/Oura.py
+++ b/scripts/artifacts/Oura.py
@@ -1,15 +1,24 @@
 __artifacts_v2__ = {
-    "oura_account_identity_profile": {
-        "name": "Oura - Account, Identity & Profile",
-        "description": "Account identity, paired-ring hardware, demographics, health "
-                       "profile, cross-service identifiers and support tokens, grouped "
-                       "by category.",
+    # The segment/zendesk/account-UID patterns below also match same-named SDK
+    # plists inside other apps' containers (Life360, etc.). Parsing is anchored
+    # to the directory that holds com.ouraring.oura.plist (_load_oura_plists),
+    # so those foreign matches are never read.
+    #
+    # Coverage architecture: every key of every Oura plist is either claimed by
+    # one of the topical artifacts below or reported raw by
+    # oura_preferences_residue - nothing is silently dropped.
+    "oura_account_identifiers": {
+        "name": "Oura - Account & Identifiers",
+        "description": "Account e-mail and UIDs plus cross-service identifiers "
+                       "and tokens: Segment user/anonymous IDs, Zendesk user id, UUID "
+                       "and auth tokens, Apple App-Attest UUID.",
         "author": "@slay3r00",
         "creation_date": "2026-06-02",
-        "last_update_date": "2026-07-31",
+        "last_update_date": "2026-08-09",
         "requirements": "none",
         "category": "Oura Ring",
-        "notes": "Merges the account/device, user-profile and identifier data into one view.",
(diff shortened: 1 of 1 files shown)


abrignoni:
@Slay3r00 thank you for jumping on this one, DaKota. The anchoring approach is the right call: requiring `com.ouraring.oura.plist` and only reading plists that sit next to it kills the false positives from #1727 at the root, and dropping the `deviceUid` fallback closes the side door too. The per-row Source File column is a great addition for provenance. Good stuff.

Three things before this can land:

1. **Rebase onto current main.** The PR is showing conflicts because `Oura.py` changed on main after you branched: the html_columns escaping fix, the sample_data population commits, and the documentation audit from July 31 all touched it. A rebase (or a fresh branch off main with your changes reapplied) will clear that up and let CI run again.

2. **Description wording.** After your branch was cut we added a CI check, `admin/scripts/check_claim_language.py`, that guards the examiner-facing `name` and `description` fields against completeness claims, since those fields end up in reports and casework. I ran your version of the file against it and it flags 7 descriptions, mostly for "every" and "all" ("every cross-service identifier", "every ring/UTC time mapping variant", "all accuracy values", "Every timestamp-bearing Oura key", "every analytics event name", "Every remaining key", plus a "habit" hit on the health profile). The fix is to say what the data is without the completeness word, for example "cross-service identifiers and tokens" instead of "every cross-service identifier and token". The parser can't promise it caught every variant a future app version might write, so the descriptions shouldn't either.

3. **sample_data.** No action needed from you on this one, just a heads up: the sample_data blocks come from my local test image corpus, and since the artifact keys are renamed and split I'll regenerate them on my end once the rebase is in. Feel free to leave those blocks out of the rebased version.

The restructure into topical artifacts plus a residue catch-all is a nice design, and I appreciate that nothing gets silently dropped. Once the rebase and rewording are in we'll get this merged.

abrignoni:
Verified against all 21 test images: the old code was reporting foreign app data on six of them (Life360, ChatGPT and others, up to 259 rows on one image), and with this fix every one of those rows is gone, no errors, and the anchor behaves as described. Removing the old sample_data entries was the right call since they recorded the foreign matches. Thank you for the quick turnaround on the report. Merging.

PR #1747, opened 2026-07-22
Fix MemoryError during HTML report generation (#1746)

### LINKED ISSUE #1746: MemoryError during HTML report generation
iLEAPP was unable to complete the processing of an iOS extraction due to a memory error whilst generating the HTML report

Image

The problem appears to have been identified: specifically, the entire HTML file generated by each artifact is being loaded in order to insert the report elements (sidebar and other controls).
This bug affects all the LEAPPs.
It also appears that this error prevented the JSON file from being generated to open the report in LAVA (to be confirmed).

JSap0914:
## Summary

Fixes #1746.

HTML report generation could crash with `MemoryError` on large extractions. In `scripts/report.py`, `generate_report()` finalizes each artifact page by reading its entire `.temphtml` file into memory (`get_file_content`) and rebuilding it with string concatenation (`insert_sidebar_code`) to inject the navigation sidebar. For large extractions a single artifact page can reach several GB, so this peaked at roughly **3x the page size** and raised `MemoryError` — matching the issue's diagnosis that "the entire HTML file generated by each artifact is being loaded in order to insert the report elements (sidebar and other controls)."

## Change

Replace the whole-file read/concat with `stream_insert_sidebar_code(src_path, dest_path, sidebar_code)`:

- Copies the page to its final `_HTML/` location in fixed-size (1 MiB) chunks.
- The sidebar placeholder (`<!--__INSERT-NAV-BAR-DATA-HERE__-->`) is written near the top of every page, so only a small head buffer is retained while searching for it. A placeholder split across a chunk boundary is handled by retaining a `len(placeholder)-1` tail.
- Once the placeholder is found, the (large) table body that follows is streamed with `shutil.copyfileobj`.
- Output is **byte-identical** to the previous approach, including the missing-placeholder fallback (page copied verbatim + logged warning).

Peak memory is now bounded by the chunk size, independent of page size.

## Reproduction / measurement

With a ~92 MB synthetic artifact page (placeholder near top + large table body):

| Approach | Peak traced memory |
|---|---|
| Before (read + concat) | ~343 MB (**3.00x** file) |
| After (streaming) | ~5 MB (**0.055x** file) |

## Tests

Adds `admin/test/scripts/test_report_streaming.py` (deterministic, no forensic data required; runnable via `pytest` or standalone):

- Byte-identical output across page shapes: large body, tiny file, placeholder at start/end, placeholder straddling a chunk boundary, empty file.
- Missing-placeholder fallback copies content verbatim.
- Bounded-memory regression: peak `< 0.5x` file size for a >16 MB page (the old approach needed ~3x).

```
$ python -m pytest admin/test/scripts/test_report_streaming.py -v
3 passed
```

This bug affects all the LEAPPs; this PR fixes the iLEAPP core report generator.


Johann-PLW:
@JSap0914 
Thank you very much for fixing this memory issue with huge HTML files.
Coud you please remove the `admin/test/scripts/test_report_streaming.py` before I merge your changes into the `scripts/report.py` file?

Johann-PLW:
@JSap0914 
Wasn't able to delete the test file form command line but worked on Github.
Thanks again for your contribution.

PR #1748, opened 2026-07-23
Add Device.Timezone Biome

CynthiavDorp:
Hi,

This script parses the Device.TimeZone Biome (`/private/var/db/biome/streams/restricted/Device.TimeZone/local/`), adding another source to see what timezone the device was in.

The Biome had only 1 other field, but my testdata didn't give me enough of an idea on how to interpret that field. If preferable, I can make it visible as a column with unknown interpretation.

Let me know if anything needs to be fixed :)


abrignoni:
Hi Cynthia,

Thank you for this contribution, and welcome! Device.TimeZone is a nice addition to the Biome coverage. The parsing logic was solid as submitted, so instead of sending it back to you I made a few small adjustments directly on the branch and I want to explain what and why.

1. **Changed `import blackboxprotobuf` to `from scripts import blackboxprotobuf`.** This one is not your fault at all. After you opened this PR we bumped protobuf to 5.29.6 for security fixes, and since the PyPI blackboxprotobuf package pins the old vulnerable protobuf, iLEAPP now ships its own vendored copy under `scripts/blackboxprotobuf` and the PyPI package is no longer installed. Your artifact was written the correct way for the iLEAPP that existed when you wrote it. Any future artifacts should use the `from scripts import blackboxprotobuf` form.

2. **Merged main into the branch** so the above actually works and the PR reflects what will land.

3. **Small cleanups:** removed two unused imports and a leftover debug print, and changed `protostuff['2']` to a guarded `.get('2')` so a single record missing the timezone field logs and skips instead of stopping the whole artifact.

Verified after the changes: the module compiles, imports, and registers correctly in the plugin loader on current main.

On the unknown field 1 you mentioned: leaving it out is fine for now. If you ever collect more test data and work out what it means, a follow-up PR would be very welcome.

Thanks again!

PR #1792, opened 2026-07-26
fix: chrome bookmarks keyerror and index bug

sandrzejewskipl:
Fixed:
- Prevent KeyError on missing bookmarks JSON fields (url, date_added, name, type) by using .get() with defaults. 
Traceback:
```
Reading chromeBookmarks artifact had errors!
Error was 'url'
File "e:\github\iLEAPP\ileapp.py", line 536, in crunch_artifacts plugin.method(files_found, category_folder, seeker, wrap_text, time_offset) ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
File "e:\github\iLEAPP\scripts\ilapfuncs.py", line 527, in wrapper data_headers, data_list, source_path = func(Context) ~~~~^^^^^^^^^ 
File "E:\github\iLEAPP\scripts\artifacts\chrome.py", line 1150, in chromeBookmarks url = valueb[index]['url'] ~~~~~~~~~~~~~^^^^^^^ KeyError: 'url'
```
- Fix bug where valueb[0] was always used instead of valueb[index], causing all bookmarks to incorrectly share the first item's name and type values.

### DIFF
--- a/scripts/artifacts/chrome.py
+++ b/scripts/artifacts/chrome.py
@@ -1147,13 +1147,13 @@ def chromeBookmarks(context):
                             if keyb == 'children':
                                 if len(valueb) > 0:
                                     for index in range(len(valueb)):
-                                        url = valueb[index]['url']
-                                        dateadd = valueb[index]['date_added']
+                                        url = valueb[index].get('url', '')
+                                        dateadd = valueb[index].get('date_added', 0)
                                         dateaddconv = datetime.datetime(1601, 1, 1) + datetime.timedelta(
                                             microseconds=int(dateadd))
                                         added_dt = dateaddconv
-                                        name = valueb[0]['name']
-                                        typed = valueb[0]['type']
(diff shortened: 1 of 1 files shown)


abrignoni:
Thank you again @sandrzejewskipl! 🙏 Merged.

I verified both fixes end-to-end. The `valueb[0]` → `valueb[index]` one is a great catch — every bookmark was silently inheriting the first entry's name and type, which is exactly the kind of quiet data bug that matters in a forensic report. And the `.get()` guard fixes the KeyError that was aborting the artifact on any bookmarks file with folders (i.e. basically all of them).

Two solid PRs in a row — really appreciate you digging in and sending clean fixes with clear tracebacks. This is a big help to the whole #DFIR community. 🚀

PR #1800, opened 2026-07-27
feat: Artifact Result Streaming

JamesHabben:
Adds ArtifactResult, a streaming result path for large artifacts.

Modules can still return the existing (headers, data_list, source_path) tuple, so current artifacts continue working unchanged. Larger artifacts can now create a result object with context.create_artifact_result(), set headers/source metadata, add rows one at a time with results.add_row(...), and return the result.

Rows added through ArtifactResult are written into the LAVA SQLite database in batches, then replayed from LAVA for HTML, TSV, timeline, and KML output. This avoids requiring modules to build one large in-memory data_list before handing results back to the core.

This PR also converts a small set of higher-volume artifacts to the new writer-style API for testing and comparison. Legacy, generator, and writer-style test runs produced matching LAVA record counts.

### Test Scripts

Temporary test/profiling scripts are included in this branch to support review of the ArtifactResult streaming work. These are not intended for final merge.

Included test scaffolding:
- `admin/test/scripts/make_perf_sqlite.py`
  - Creates a synthetic SQLite fixture with narrow, medium, and wide 1M-row artifact-style datasets.
  - Used to compare behavior across small rows, moderate transformed rows, and very wide rows.

- `admin/test/scripts/benchmark_perf_artifacts.py`
  - Runs selected performance profiles and captures processing time, wall time, and memory use.
  - Used to compare legacy list-return modules against ArtifactResult-backed modules.

- `admin/test/perf_profiles/*.ilprofile`
  - Temporary profiles for running narrow, medium, wide, sync/async, and conversion comparison scenarios.

- `admin/test/perf_data/artifact_result_fixture/.../leapp_perf_artifact_result.sqlite`
  - Generated synthetic fixture database used by the benchmark profiles.

These files were included only so reviewers can reproduce the measurements during feature testing. The plan is to remove this temporary benchmarking scaffolding before the final merge.

### DIFF
--- a/scripts/lavafuncs.py
+++ b/scripts/lavafuncs.py
@@ -18,6 +18,7 @@ Functions:
     lava_add_module: Adds module information to the LAVA data.
     lava_create_sqlite_table: Creates a SQLite table for artifact data.
     lava_insert_sqlite_data: Inserts data rows into a SQLite table.
+    lava_iter_artifact_rows: Streams stored artifact rows back from LAVA.
     lava_get_media_item: Retrieves media item information from database.
     lava_insert_sqlite_media_item: Inserts media item metadata into database.
     lava_get_media_references: Retrieves media reference information.
@@ -34,6 +35,8 @@ from platform import platform
 from collections import OrderedDict
 import re
 import datetime
+import queue
+import threading
 
 from scripts.version_info import leapp_name, leapp_version
 from scripts.context import Context
@@ -43,6 +46,8 @@ lava_data = None
 lava_db = None
 lava_db_name = '_lava_artifacts.db'
 lava_json_name = '_lava_data.lava'
+lava_db_path = None
+_QUEUE_STOP = object()
 
 
 def sanitize_sql_name(name):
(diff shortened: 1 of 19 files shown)


scripts/lavafuncs.py
```diff
@@ -413,40 +565,109 @@ def lava_insert_sqlite_data(table_name, data, object_columns, headers, column_ma
...
+                state["inserted_count"] += len(prepared_batch)
+
+            db.commit()
+        except (sqlite3.Error, TypeError, ValueError) as ex:
+            state["error"] = ex
+            db.rollback()
+        finally:
+            db.close()
+
+    thread = threading.Thread(target=writer, name="LEAPPLavaArtifactWriter", daemon=True)
```
bconstanzo, line 637: Just as a comment, I've grown very fond of concurrent.futures.ThreadPoolExecutor() to not worry about threading. Not sure if it'd work here, but sometimes can make it simpler.

abrignoni:
Pushed the merge with main and three fixes on top of it, all on this branch.

- A module that raises after streaming rows now leaves nothing behind: the table is dropped and the manifest entry removed, the same as a list-returning module that raises. Before, 80,000 of 84,972 rows stayed under an entry with no record count.
- Replayed rows carry the values the module wrote. Floats, booleans and integers in untyped columns round-trip, so the TSV, timeline and KML of a converted artifact come out identical to the list path's. The list path's own storage is unchanged, including the whole-second int epoch for an ISO string, which the unified handler had turned into a float.
- fileSystemEvents streams. FSEvents Location Services on the iPhone 11 image, 3,569,449 rows: peak RSS 2,982 MB as a list, 1,214 MB streamed, same core, with the LAVA table, manifest, TSV and HTML page identical.

Proofs, all on a real iOS 17 image. Streaming against the same ten artifacts restored to list form on the same core: 18 LAVA tables, 18 TSVs, the KML and the timeline identical. Unconverted artifacts against main on a 34-artifact profile: 28 of 28 tables and TSVs identical, timeline identical. 561 unit tests pass and lint_changed reports no new warnings.

The description now matches what is in the branch. The perf scaffolding is still here for you to take out when you are ready.

PR #1801, opened 2026-07-28
New artifacts: Biome Networking Edge Selection + DuetExpertCenter notifications

abrignoni:
Two new iOS artifacts, developed from the paths and validated end-to-end against real corpus extractions.

## Biome - Networking Edge Selection
`*/streams/*/Device.Networking.EdgeSelection/local/*`

Each SEGB record is a network-edge observation. Decoded fields:

| Column | Example |
|---|---|
| SEGB Timestamp | 2024-08-21 22:36:38 UTC |
| IP Address | `2600:380:6c59:ad00::` / `69.143.130.0` |
| IP Version / Prefix Length | 6 / 56 · 4 / 24 |
| Interface | `Cellular` · `WiFi` |
| Radio Technology | `LTE` · `802.11ax` · `802.11ac` |
| Country / Time Zone | `US` / `America/New_York` |

Forensically this places the device on a specific carrier or Wi-Fi network at a point in time — the public address is the endpoint the device saw for *itself*. The Otto sample even shows a time-zone shift to `America/Denver` on cellular, a travel indicator.

Tombstone files carry deletion bookkeeping under a **different** schema (source filename, offsets, `biomed`), so they're skipped.

## DuetExpertCenter - Notifications / Notification Suggestions
`*/mobile/Library/DuetExpertCenter/notificationAndSuggestionDB.db*`

The notification history iOS keeps to rank/summarize notifications: delivering app, urgency, delivery method + reason, prominence, expansions, thread/contact ids, and the user's latest interaction outcome with its timestamp. Plus the Smart Pause style suggestions it generates, joinable back via `Trigger Notification UUID`.

**Why it matters:** this DB frequently retains far more history than an app's own storage, so it can evidence app activity after the app's data is gone (12,396 notification records on one image). It records *that* a notification arrived and how it was handled — not its content. Apple-internal numeric codes (urgency, delivery method/reason, outcome) are reported **verbatim** rather than guessed at.

**Schema-drift guard:** `notificationBodyLength` is absent on some iOS 17 builds. Naming it unconditionally raised `no such column` and silently dropped **every** row — 0 instead of 3,273 on one corpus image. Late-added columns are now selected only when present.

## Validation (6 corpus extractions, iOS 16.5 – 18.7)

| Image | Edge | Notifications | Suggestions |
|---|---:|---:|---:|
| Otto 17.5.1 | 8 | 12,396 | 100 |
| iPhone 11 17.3 | 0 | 3,273 | 11 |
| Dexter 18.3.2 | 0 | 3,109 | 14 |
| HC 18.7.8 | 0 | 171 | 117 |
| Felix 16.5 | 0 | 120 | 0 |
| fs-full-002 17.1 | 0 | 27 | 0 |

Every image parses without error. The 0-row Edge results were verified as genuinely empty (HC's stream files hold 7 records with **0 written**; the others have empty stream directories) — not a parser gap. `sample_data` recorded for each.

pylint `--disable=C,R` = **10.00** · `py_compile` clean · PluginLoader 725 → **728** with all three registered · icons verified against `tabler-icons.css`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/artifacts/biomeNetworkingEdgeSelection.py
+++ b/scripts/artifacts/biomeNetworkingEdgeSelection.py
@@ -1,9 +1,9 @@
 __artifacts_v2__ = {
     "get_biomeNetworkingEdgeSelection": {
         "name": "Biome - Networking Edge Selection",
-        "description": "Parses the public-facing IP address, interface type, radio technology, "
-                       "country and device time zone recorded by the Device.Networking.EdgeSelection "
-                       "biome stream",
+        "description": "Parses the public-facing network prefix (a truncated IP address), interface "
+                       "type, radio technology, country and device time zone recorded by the "
+                       "Device.Networking.EdgeSelection biome stream",
         "author": "@AlexisBrignoni, Claude",
         "creation_date": "2026-07-28",
         "last_update_date": "2026-07-28",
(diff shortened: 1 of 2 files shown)


abrignoni:
**Update — column-name provenance tightened on the Edge Selection artifact (c730dc7).**

Protobuf carries field *numbers* but no field *names*, so every column here except the SEGB timestamp is an inference from observed values. Three corrections so the report can't be over-read:

1. **The address is truncated.** All 8 observed records have their host bits zeroed relative to field 3 (verified with `ipaddress.ip_network`): `69.143.130.0` is the **/24 network**, `2600:380:1871:6d00::` the **/56** — not the device's full public IP. Column renamed **`IP Address (Truncated)`**, with the description and notes stating it identifies a *network*, so it isn't mistaken for an endpoint address in a report or affidavit. (This also corroborates the "Prefix Length" reading of field 3 — it exactly explains the zeroing.)
2. **Field 6 → `Field 6`.** Values seen are `Unknown` / `None` / empty with no established meaning, so the invented "Additional Info" label is gone, matching the convention already used in `biomeNotificationsPub`.
3. **Notes now state** the remaining column names are inferred from observed values.

For contrast, the **DuetExpertCenter** column names are Apple's own SQLite column names (lightly prettified) and are verifiable against `.schema notifications` — no inference there.

No parsing change: same 8 rows on Otto, pylint `--disable=C,R` = 10.00.

PR #1806, opened 2026-07-28
Update FitnessWorkoutsLocationData.py

MetadataForensics:
Update for column name changes from iOS 17 to iOS 18 and later.


abrignoni:
Thank you James. I updated the branch onto current main, kept your generation branch over the single query that was there (yours preserves the workout linkage where the newer association names exist), and pointed the column check at the database the artifact already found. Verified against all 21 test images: the one image with workout location data reproduces exactly (20 and 25,473 rows), no errors anywhere. None of our images carry the newer schema, so that path rests on your data. Merging. Good stuff.

MetadataForensics:
Ran off my newer schema data, got the following row counts:
Screenshot 2026-08-10 155903
Thanks very much!

PR #1811, opened 2026-07-28
fix: catch edge case where full host file path was stored with media

JamesHabben:


abrignoni:
Thanks for this, @JamesHabben. Closing it because the fix landed by another route in #2215, which records evidence-relative source paths across all five seekers rather than two.

Checked before closing, on a directory input at current main: 66 media rows, none carrying a host path. Reverting the one line #2215 changed in `FileSeekerDir` puts the full report path on all 66, so that guard is what is doing the work.

The `get_relative_path` change here would also not apply cleanly now. It goes back to a `startswith` prefix strip, and main removes the base wherever it appears in the string.

One idea from this PR is still worth having on its own. `_check_in_media` stores whatever the seeker recorded without scrubbing it, and running that through `get_relative_path` first would have contained the leak even with #2215 reverted. Happy to take that as a small PR if you want to open one.

Issue #1227, opened 2025-08-08
[Feature Request] Runtime key command to skip an artifact


sharuk998:
While running iLEAPP, sometimes a specific artifact may take a long time or appear to hang on certain artifact. Currently, the only way to skip it is to rerun the entire process without that plugin specified in the ilprofile file.

It would be very useful if, while iLEAPP is running, the user could press a certain key (e.g., s) in the terminal to instruct the program to:
- Stop processing the current artifact.
- Move on to the next one without terminating the whole run.

**Example usage:**
```
[152/308] mapsSync [mapsSync] artifact started
Press "s" to skip this artifact
```
When the user presses s, the program prints:
`Skipping mapSync...`
and continues with the next artifact

**Benefits:**
- Saves time when a single plugin hangs.
- No need to restart the run.
- Helps troubleshoot which plugins are problematic on a given backup.

JamesHabben:
Some initial thoughts. In order to do this we have to split the work. Everything is currently done in a single linear process. We will need to multi-process in order to have a monitor for keystroke and the ability to kill off other work. Not a major hurdle but will need a careful refactor plan to implement.

Mudassir2021:
Hi, are we going to add this feature soon? We are using this package for [SUMURI Recon System](https://sumuri.com/software/recon-itr/?srsltid=AfmBOoqJuIJ55G23MuRwfFwLtWHvXUL34IxOrarUnT4Avm6FDuzhupkV) iOS Backup Triage tool, and the functionality is incomplete without having the ability to skip the plugin in case any of them gets stuck...

abrignoni:
Currently we are not working on refactoring the code for the multiprocess feature but the request has started conversations on how to do so.

That being said as an open-source, unpaid volunteer led project we are open to receiving pull requests on any and all improvements to the codebase. 🙂

Issue #1226, opened 2025-08-08
[Bug] mapSync artifact causes iLEAPP to hang on encrypted iOS backup


sharuk998:
While running iLEAPP on a decrypted iOS backup (originally encrypted), the mapSync plugin caused the process to hang indefinitely. When I re-ran iLEAPP excluding the mapSync artifact, it completed successfully.

**Steps to Reproduce:**
- Decrypt encrypted iOS backup using [jfarley248/iTunes_Backup_Reader](https://github.com/jfarley248/iTunes_Backup_Reader).
- Run iLEAPP on the decrypted output folder with all artifacts selected.
- The run gets stuck on mapSync and never completes.

**Expected behaviour:**
The tool should either complete processing or provide an error message rather than hanging.

**Observed behavior:**
The process stays stuck at mapSync indefinitely (no further logs, CPU usage stays constant).

**Environment:**
OS: MacOS 15.5 (24F74)
Python version: 3.12

**Command used:**
`python ileapp.py -t itunes -i "/path/to/decrypted_backup" -o /path/to/output`

**Log file:**
I’ve attached the _Script_Logs folder from the run that got stuck.
[_Script_Logs.zip](https://github.com/user-attachments/files/21681196/_Script_Logs.zip)

JamesHabben:
Do you have sample data to share or know of a public test image that can reproduce the hang issue?

abrignoni:
@sharuk998 thank you for sticking with this, and more importantly for posting the backup. That upload is the only reason any of this could be checked, and it is appreciated.

I could not reproduce the hang. Here is exactly what I ran, so you can tell me where our setups differ.

Using your `00008110-001A08AA3CEB801E-Encrypted.zip` with the password you gave, all artifacts selected, `-t itunes` with iLEAPP's built in decryptor, which is the path you described in #1248:

- **Current main (`94f93173`): 973 artifacts, 13 seconds wall clock.** mapsSync ran at 544/973 and returned 3 records, no error.
- **A checkout from the day you filed #1248 (`d49bf2e2`, 2025-09-05): 437 artifacts, 5 seconds.** mapsSync did fail there, but it raised `IndexError: list index out of range` at `mapsSync.py` line 154. iLEAPP caught it, logged it, and the run carried on to the end.

[...]

PR #1829, opened 2026-07-30
Feature skip plugins mid run

Abu-Huraira21:
Adds --mp_per_plugin, an opt-in flag that runs each plugin in its own subprocess instead of the main process. Default
  behavior (no flag) is unchanged.

  Problem: a single hung/stuck plugin blocks the entire run with no way to recover short of killing the process.

  Fix: with --mp_per_plugin, each plugin runs in a spawned subprocess. If one hangs:
  - Ctrl+C once → skips just that plugin, run continues
  - Ctrl+C twice within 2s → aborts the whole run
  - kill -USR1 <pid> / kill -USR2 <pid> (Unix) → skips current plugin externally, no abort risk

  Files changed

  - ileapp.py — --mp_per_plugin flag, subprocess dispatch, skip signal handlers
  - scripts/mp_plugin_runner.py (new) — subprocess entrypoint
  - scripts/lavafuncs.py — idempotent LAVA init, open/close helpers for subprocess-safe DB access
  - scripts/ilapfuncs.py — picklable seeker/file-info snapshot for passing state into subprocesses

  Known issue (not fixed by this PR)

  Encrypted iTunes backups can fail to open the decrypted Manifest.db with unable to open database file — pre-existing,
  unrelated to this change. Root cause: Manifest.db is WAL-mode, and open_sqlite_db_readonly() has no fallback for opening
  WAL-mode DBs missing their -wal/-shm sidecar files. Reproducible on current main independent of --mp_per_plugin. Will
  follow up with a separate PR (immutable=1 retry fix).

  Testing

  Verified --mp_per_plugin end-to-end against an iTunes backup extraction, including plugin skip via Ctrl+C and SIGUSR1.
  

Notes:
 - This PR was made against this request https://github.com/abrignoni/iLEAPP/issues/1227

### DIFF
--- a/ileapp.py
+++ b/ileapp.py
@@ -217,7 +217,23 @@ def main():
         parser.print_help(sys.stderr)
         sys.exit()
 
-    args = parser.parse_args()
+    # Filter out multiprocessing spawn arguments that argparse doesn't recognize
+    # When using multiprocessing with 'spawn', Python may inject internal flags that
+    # our CLI parser doesn't know about (and would otherwise treat as fatal errors).
+    filtered_argv = []
+    for arg in sys.argv[1:]:
+        # Skip multiprocessing / interpreter-injected arguments
+        if (
+            arg.startswith('--multiprocessing-')
+            or arg.startswith('tracker_fd=')
+            or arg.startswith('pipe_handle=')
+            or arg in ['-B', '-S', '-I']  # Python optimization / startup flags
+        ):
+            continue
+        filtered_argv.append(arg)
+
+    # Use filtered arguments for parsing so frozen mp children don't break argparse
+    args = parser.parse_args(filtered_argv)
 
     available_plugins = []
     loader_paths = [plugin_loader.PLUGINPATH]
@@ -792,4 +808,6 @@ def crunch_artifacts(
     return True
 
 if __name__ == '__main__':
+    # Support for multiprocessing with PyInstaller frozen binaries
+    multiprocessing.freeze_support()
     main()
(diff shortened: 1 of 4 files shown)


abrignoni:
Thanks for this, @Abu-Huraira21, and for the second pass in ALEAPP. It works: run against a deliberately hung artifact on a zip input, SIGUSR1 skipped it, the run continued, and the report came out identical to a run without the flag on 14 artifacts.

Closing it anyway, for three reasons.

The case that opened #1227 did not reproduce. Running the reporter's own backup (details in #1226) showed a crash in older code, fixed since, not a hang. What shipped instead is a start time and elapsed seconds for every artifact, in all five LEAPPs (#2062 here), so a slow artifact is now visible in the log.

The subprocess model leaks run-level state. A child sets `identifiers`, the keychain path and the iOS version in module globals and only the version comes back, so the Device Details page comes out empty and keychain-backed artifacts report their data as still encrypted. `artifact_processor_streaming` artifacts fail on `len()` of an iterator and their rows land in the database with no manifest entry. `seeker.search()` from a child returns nothing on zip or tar input. And a skipped artifact is not recorded anywhere an examiner reads, while the log prints "artifact completed" for it.

#1800 rewrites the same three files and is going in first.

The isolation is worth having, for crash containment and per-artifact memory as much as for skip. It will be rebuilt as one design across the five cores on top of #1800, with a worker that owns the run state so nothing has to be marshalled back by hand. Your two PRs are the reference for that work.

PR #1838, opened 2026-07-30
CacheDelete Purge History

Jadoo4QFan:
This file, which is in almost every extraction, tracks the history of the internal CacheDelete (deleted) daemon with logs going years, which tracks the history of cache being purged, which can be useful when destruction of evidence is suspected or when cache is sparse.


abrignoni:
Thank you for this one. I pushed a fixup to your branch: the metadata dict had an extra closing brace so the module never imported (that was the CI failure), the notes now state the timestamp zone is undocumented and that the urgency mapping comes from the Apple Wiki, and every matched copy of the file is parsed with a Source File column instead of the first only. Ran it against our 21 test images: none carry this file, so the run is clean but the parser rests on your test data until a corpus image with purge history shows up. One for the test data wish list. Merging.

PR #1889, opened 2026-08-07
Added Claude Artifacts

bcbtree:
Added parsers for conversations, messages, and projects, as well as account info for profile on claude application.


stark4n6:
@bcbtree same as over on ALEAPP, do you think you can combine them into one .py?

bcbtree:
@stark4n6 Updated to one .py

abrignoni:
Thank you for these, Brandon. Same maintainer fixup as the ALEAPP side before merging: last_update_date field name, punctuation in the notes, a LEFT join for completeness, and sample_data from validation runs on two of our test images (1 account, 2 conversations, 10 messages on iOS 18.7; 1/7/30 and 1 project on iOS 26, matching the store row for row). I also updated the branch onto current main and made the account artifact pick the bootstrap json that actually holds the account object, since iOS container paths cannot anchor the glob to the app. Good stuff.

PR #1919, opened 2026-08-09
fix: strip rollback journals alongside the other sqlite sidecars

abrignoni:
Artifacts that strip SQLite sidecars from their matched-file list enumerate the suffixes, and most of them list only `wal` and `shm`. A rollback journal survives that filter:

```python
files_found = [x for x in files_found if not x.endswith('wal') and not x.endswith('shm')]
```

Given `foo.db`, `foo.db-wal`, `foo.db-shm`, `foo.db-journal`, the filter returns the database **and the journal**. Where the artifact then takes `files_found[0]`, whichever the seeker returned first is opened as a database.

`journal` is now stripped alongside the other two, completing a set that two lines in the codebase already had.

## This is completeness, not a bug fix

I checked whether the hazard actually occurs before proposing the change. Every glob from every affected module was run against all registered zip corpora, iOS and Android: **zero cases where a `-journal` file survives the filter.** Nothing currently misbehaves because of this.

It is worth closing anyway because the state is entirely reachable, unlike genuinely dead code. The rollback journal is SQLite's *default*-mode sidecar, journals do appear in real extractions, and these globs end in a wildcard that would match one. Our corpora simply happen not to pair a hot journal with these particular databases.

## Why suffix stripping rather than selecting the database by name

Selecting positively, `x.endswith('<the database>')`, is the better idiom and is what the rest of the codebase does. It was rejected here as the larger change: 22 modules would each need their real database name looked up, it alters *which* file is chosen so every one needs corpus verification, and it lands immediately before the core consolidation. Adding one suffix changes no selection behaviour and has nothing to regress.

## Scope

| core | lines | modules |
|---|---|---|
| iLEAPP | 9 | 2 |
| ALEAPP | 32 | 18 |
| DLEAPP, RLEAPP, VLEAPP | 0 | none — they select databases by name already |

Verified by AST across both cores: no comprehension filters `wal` without also filtering `journal`. Behaviour confirmed directly — given a database plus all three sidecars, the filter now returns the database alone.

`check_claim_language.py`, `check_html_safety.py` and `lint_changed.py` pass with no new warnings.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/scripts/artifacts/Airbnb.py
+++ b/scripts/artifacts/Airbnb.py
@@ -39,10 +39,15 @@ from scripts.ilapfuncs import artifact_processor, get_sqlite_db_records, convert
 
 @artifact_processor
 def airbnb_messages(context):
-    files_found = [x for x in context.get_files_found() if not x.endswith('wal') and not x.endswith('shm')]
-    
+    files_found = [x for x in context.get_files_found() if x.endswith('.sqlite3')]
+
+    # The account id is embedded in the database filename as _<digits>_. A build
+    # that names it differently is not a reason to lose the messages, so the id
+    # is reported empty rather than raising out of the artifact: the exception
+    # would be an AttributeError, which no except sqlite3.Error here would catch.
     stem = Path(files_found[0]).stem
-    user_account_id = int(re.search(r'(?<=_)\d+(?=_)', stem).group())
+    account_match = re.search(r'(?<=_)\d+(?=_)', stem)
+    user_account_id = int(account_match.group()) if account_match else ''
 
     query = ('''
         SELECT
--- a/scripts/artifacts/WithingsHealthMate.py
+++ b/scripts/artifacts/WithingsHealthMate.py
@@ -153,7 +153,8 @@ def get_healthmate_accounts(context):
 @artifact_processor
 def get_healthmate_sleep_tracking(context):
     files_found = context.get_files_found()
-    files_found = [x for x in files_found if not x.endswith('wal') and not x.endswith('shm')]
+    files_found = [x for x in files_found if not x.endswith('wal') and not x.endswith('shm')
+                   and not x.endswith('journal')]
     query = ('''
         SELECT
         ZDEVICEID,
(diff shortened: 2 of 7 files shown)


abrignoni:
Correction to this PR's description, after the fact.

This PR was branched from a working tree that already carried five unrelated, unfinished artifact fixes, and `git add scripts/artifacts/` staged all of them. So it merged more than the title and body describe:

**Journal filter, as described:** `ZangiMessenger.py`, `WithingsHealthMate.py`.

**Not described, and merged inadvertently:**
- `splitwise.py` — selects its database by schema rather than taking the first `database.sqlite` the glob matches (the glob has no app-specific component, so it was opening other apps' databases)
- `whatsApp.py` — guards `ZWACDCALLEVENT.ZGROUPCALLCREATORUSERJIDSTRING`, absent on older releases
- `photosMigration.py`, `trustedPeers.py` — existence guards for tables absent on some releases
- `Airbnb.py` — selects the database by name, and guards a `re.search(...).group()` that raises `AttributeError` when the filename carries no `_<digits>_`

Those five were meant to go through a 21-corpus regression sweep before shipping and did not. They pass the CI gates and byte-compile, but they have not been verified against the corpora. A clean sweep from current main is running now and I will post the result here; anything that regresses will be reverted.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

abrignoni:
Verification result for the five files that merged here undescribed.

Targeted profile of the 22 artifacts in all 7 changed modules, run from a worktree pinned at this merge commit, against all 21 registered iOS corpora, compared to the pre-merge baseline:

| | |
|---|---|
| artifact/corpus pairs identical | **84** |
| gains | 0 |
| **regressions** | **0** |
| error lines | **0 on every corpus** |

Zero gains is the expected shape: these were error-elimination fixes rather than recovery fixes. `splitwise` was opening another app's `database.sqlite` on corpora that carry no Splitwise data, and `photosMigration` and `trustedPeers` guard tables that are genuinely absent on those releases. Errors went to zero while row counts stayed identical, which is what a correct guard looks like.

Nothing regressed, so nothing is being reverted.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

PR #1934, opened 2026-08-11
Lidl Plus Forensics: SQLCipher, Keychain & LAVA

djangofaiola:
**Supported Artifacts:**

📋 Shopping List
📄 Ticket List
🧾 Receipt Items
💳 Payment Methods
🎴 Loyalty Card
🏷️ Promotion Details
📦 Product Details
🔍 Searched Terms
🏪 Store Searches
🎟️ Offers
🎁 Loyalty Points Marketplace
📍 Last Known Location
🛒 Self Scanning Basket
📲 Self Scanning Journey
🗑️ Removed Self Scanning Items
🧺 Grocery Pickup Cart
👤 Account


abrignoni (commented):
Thanks for this one, it is a big piece of work and the forensic hygiene shows. The Kingfisher matching hashes the URL and looks it up against the cache filename instead of guessing by size or mtime, the keychain lookup refuses an ambiguous match instead of picking one, and the decryption records HMAC verified/partial/failed per database. I also checked all five cores: there is no Lidl parser anywhere, so this is new coverage with no leveling needed.

A few things to fix before this goes in.

**1. The remote links have to go. This is why CI is red.**

`format_url()` builds `<a href="{url}" target="_blank">` at line 644 and `admin/scripts/check_html_safety.py` flags it. That check exists so a report never becomes a beacon. An examiner opening the HTML should not be reaching out to a third party server. Its BASELINE is empty, so this is the first new violation since the check landed.

You lose nothing by dropping the anchor. `check_in_media()` already puts a local thumbnail in the report, and the plain text branch already renders the URL.

**2. `Line Item` gets raw server HTML with no escaping.**

`lidl_receipt_items` declares `html_columns: ["Line Item"]` and writes `article['content']` into it, which is the inner HTML of a span straight out of the cached `htmlPrintedReceipt`. Columns in `html_columns` are written verbatim, see `artifact_report.py:110`. The static checker misses it because the module references an escaper elsewhere inside `format_url`, and that satisfies the rule module wide.

**3. Five of your six HTML rows never render.**

`ilapfuncs.py:561-564` rebinds `html_data_list` whenever an artifact declares a `media` header, so the `data_list_html` you return gets discarded. That hits shopping_list, promotion_details, product_details, offers and mypoints. Only `lidl_grocery_pickup_cart` has no media column, so it is the single artifact whose anchor actually reaches the report.

The catch runs the other way too. Those five keep `"Image URL"` in `html_columns`, and what renders there is the plain `format_url()` return, which is the raw string with no escaping.

Items 1, 2 and 3 are really one fix. Drop the anchors, have `format_url` return escaped text, and remove the second `data_list_html` list from the five media artifacts.

**4. `artifact_icon: "scan-line"` does not resolve.**

I checked it against all 5,147 names in `scripts/_elements/tabler-icons.css` and the 103 in `scripts/data/feather_to_tabler_icon_names.json`. It is in neither, so `report.py` falls back to `alert-triangle` and Self Scanning Journey ships with a warning triangle in the sidebar. `scan` or `barcode` both exist. The other 13 icons are fine.

**5. No `sample_data` on any of the 17 artifacts.** Please add row counts so there is something to reproduce against.

**6. Birthdate.** `('Birthdate (Stored UTC)', 'datetime')` passes the plist value through raw, unlike every other timestamp in the module, and "(Stored UTC)" claims a timezone the code never checks. `date` is a supported type and fits a birthdate better.

Small stuff: `SELFSCAN_DB_KEYS` at line 307 is dead, `DB_CONFIGS` already carries the same key.

One to talk about rather than change: you write decrypted databases into the report folder, where sessionIOS and signalIOS use the system temp directory. After the temp leaks we just fixed, yours may well be the better call. I just want that to be a decision and not a difference.

Good stuff. Fix those and let's get it merged.

djangofaiola:
@abrignoni 

> One to talk about rather than change: you write decrypted databases into the report folder, where sessionIOS and signalIOS use the system temp directory. After the temp leaks we just fixed, yours may well be the better call. I just want that to be a decision and not a difference.

Yes, this is intentional. I consider the decrypted database a derived forensic artifact, not temporary working data. I therefore keep it in a dedicated Decrypted Databases directory at the root of the iLEAPP output, preserving the original relative path. That keeps it associated with the case, avoids temp-file cleanup/leak issues, and makes the decrypted copy available for later validation and reproducibility.

I also made the output location an explicit choice. Rather than placing the decrypted copy alongside the encrypted source database, I store derived databases under a dedicated Decrypted Databases directory at the root of the iLEAPP output, while preserving the original relative source path underneath it. This keeps original evidence and generated artifacts clearly separated, while still making the relationship between the encrypted source and its decrypted derivative unambiguous.

Thanks for the review and for pointing this out.

abrignoni:
Merged, thanks. Nice work on this one.

On the decrypted databases, your reasoning holds and I am taking it as the deliberate choice.

One thing I changed on the way in. `lidl_store_searches` had `output_types: ["all"]` in your first commit and `["standard"]` in the second, and it still declares Latitude and Longitude, so it quietly stopped writing a KML. Those are the request coordinates out of the query string, which is the device's own position at the time of the search, so that is location evidence worth plotting. I put it back to `["all"]` in https://github.com/abrignoni/iLEAPP/pull/1941. Last Known Location was unaffected and keeps its KML.

Worth knowing when you look at the output: the artifact emits one row per returned store, so the request coordinate repeats across every result row for a single search and you get stacked pins on the same spot. Not a problem, just so it does not read as a bug. If you would rather collapse those to one point per search, that is a follow-up whenever you feel like it.

Thanks again.

Issue #1649, opened 2026-07-03
Conversation view coverage: remaining chat artifacts (test data needed)


abrignoni:
We recently added LAVA threaded conversation views across the chat artifacts (see #1645, #1647, #1648). Every direction column we shipped was verified against real test data before merging. A wrong `directionSentValue` doesn't crash anything, it silently flips who said what, and that is not acceptable for a forensic tool. So the artifacts below are on hold until we can verify them the same way.

Here is what is left and what each one needs:

- [ ] **Vipps** (`vipps.py`) - The direction value comes straight from the app's JSON and we don't know its vocabulary (could be `INCOMING`/`OUTGOING`, `IN`/`OUT`, something else). Needs a Vipps extraction where the examiner knows which messages were sent and which were received.
- [ ] **Pinger Text Free** (`pingertextfree.py`) - Same situation. `ZDIRECTION` is emitted raw from the database and the value set is unverified. Needs a Text Free extraction with known sent/received messages.
- [x] **IMO HD** (`imoHD_Chat.py`) - Direction is already verified (`Message Status` = Sent/Received), but there is no thread column on sent rows. The fix is plumbing the `buid` through the query as a Chat Partner column. This one is a dev task, test data exists.
- [ ] **BeReal** (`BeReal.py`) - Textbook fit for the view (Thread ID, Direction, Sender all present). On hold only to avoid conflicting with the open BeReal rework in #1637. Once that lands, the view is a metadata-only addition.

**How to help:** if you have (or can generate) a test extraction for any of these apps where you know the ground truth of who sent what, drop a note here. The validation itself is quick: we run the artifact against your data and confirm the sent/received split matches what you know to be true. That's it.

If you want to see the pattern these conversions follow, #1647 and #1648 are good examples, including how we derive direction from account owner data already inside the evidence.

JamesHabben:
there is overlap with #1254

Issue #1254, opened 2025-09-13
Adapt chat application parsers for the LAVA Thread/Communication/Chat view


prosch88:
LAVA offers the option of displaying parsed chats in a conversation view (comparable to the actual view on the device).

Not all parsers currently make use of this option.

The following parsers are currently compatible:

- LinkedIn
- PotatoChat 
- Kleinanzeigen
- SMS
- BeReal
- Burner
- BurnerCache
- Teams

The following parsers require adjustments for the chat view:

- Oops
- ZangiChats
- bumble
- DiscordChats
- DiscordChatsA
- facebookMessenger
- imoHD_Chat
- kijijiConversations (This app uses the same structure as Kleinanzeigen)
- kikMessages
- Life360
- Line
- pingertextfree
- Slack
- telegramMessages
- teleguard
- tikTok
- Viber
- WhatsApp 

Artifacts that already use the "__artifacts_v2__" structure are mostly easy to adapt.

BrunoFischerGermany:
I think a chat view would be very helpful for investigators!
That way, you can follow a conversation chronologically.

JamesHabben:
> I think a chat view would be very helpful for investigators!
> That way, you can follow a conversation chronologically. 

Absolutely. We have one already built and getting some final tweaks. Almost ready to launch LAVA.

![image](https://github.com/user-attachments/assets/15c85122-3aac-460a-a299-db501e8082de)

JamesHabben:
LAVA is updated to use 'conversation' terms now. its backwards compatible with the previous terms for now, but we intend to remove those soon.

i also updated the artifact info docs with the latest names for everything.
https://github.com/abrignoni/iLEAPP/blob/main/admin/docs/artifact_info_block.md

JamesHabben:
updated top comment to be a checklist and checked off those already converted to current conversation form. still a few left.

PR #1938, opened 2026-08-11
feature: imo chat conversation view and key value store

JamesHabben:
partially addresses both #1649 and #1254


alanistorres2026-art (commented):
Ver

PR #1982, opened 2026-08-17
ilapfuncs updates

stark4n6:
- Deprecate  and remove logdevinfo
- Update device_info to truncate source if needed


ileapp.py
stark4n6, line 1: @abrignoni this was the pesky item, I don't think it's needed but maybe I'm wrong

abrignoni:
Ran this against jess_ios15 with a profile of the affected artifacts. Row counts match the pre-merge main and the Device Info page looks right: the source tooltips are the relative extraction path now instead of the local output folder, and the wipe indicators land in their own section instead of loose lines at the top. Nice cleanup.

One thing this broke on the way in. `admin/test/scripts/test_module.py` line 183 patched the removed function without `create=True`:

```python
patch('scripts.ilapfuncs.logdevinfo', mock_logdevinfo),
```

That threw `AttributeError: ... does not have the attribute 'logdevinfo'`, so the module test harness failed before running any artifact. CI did not catch it because the file has no TestCase, so unittest discover just imports it and moves on. Already fixed on main in 86d12c3a.

Two smaller things:

- Nothing reads `apple_device_region_code_to_region.json` yet. When the lookup gets wired up, keep in mind eight of the keys are a single letter and `lookup_metadata` is a flat get, so it has to match the longest suffix first or a code like `TA` resolves on `T`. Where did the mapping come from? Good to have the source on hand before it shows up in a report.
- The docs still described logdevinfo as being deprecated and the generator still scanned for it, so the generated table came back empty. Cleaned up in #1990.

Thanks Kevin.

stark4n6:
@abrignoni region code usage is coming in a future parser, will test it further to make sure it's proper

PR #1983, opened 2026-08-17
feat: add Destination Caller ID column to SMS artifact

vikrosj:
message.destination_caller_id records the local phone number the device used for a given message transaction (e.g. useful on dual-SIM or multi-identity devices). Account/Account Login fields is still kept. Uses  destination_caller_id to label sent messages in the threaded conversation view, replacing the existing message.account value.

For reference to destination_caller_id see [thebinaryhick](https://thebinaryhick.blog/2022/12/06/mo-sims-mo-problems-examining-phones-with-dual-sims/)


abrignoni:
Thanks for the contribution!

The changed artifact modules cite public research images in their `sample_data`, so a maintainer can generate the test fixtures from those images. Nothing is needed from you, though you are welcome to add the fixtures yourself with `admin/test/scripts/make_test_data.py`.

- `sms.py`: cites `abe_ios16`, `ctf2020_ios12`, `dexter_ios18`, `felix23_ios16`, `felix_ios17`, `hickman_ios13`, `hickman_ios14`, `iphone11_ios17`, `iphone12_ios18`, `iphone14plus_ios18`, `magnet_ios16`, `otto_ios17`; a maintainer can generate the fixture.

abrignoni:
Thanks for adding this, the destination caller ID is a good column to surface.

Two things before this can merge.

The DataFrame column list in the chat-view block got reordered when the new column went in, but the data_list tuples kept their original order, so 20 of the 22 positions now carry the wrong label. Running the module against the committed sms test case (josh_ios15_ffs) crashes with KeyError: 'image.000000.jpg' in chat_rendering.py, because the tuple's attachment filename lands under content-type. The report table itself is fine, data_headers and the tuple still agree. The fix is one list: keep the original column order and insert 'Destination Caller ID' at position 15, right after 'Account Login', which is where record[22] sits in the tuple.

Second, the new column changes the recorded output shape, so the sms baseline needs re-recording in the same PR once the fix is in: run python admin/test/scripts/test_module.py sms -a all -c all, commit the new snapshot in admin/test/results/sms/ and delete the superseded one.

The test case data is already in the repo, so the crash reproduces with the command above.

vikrosj:
Thanks for your reply! I haven´t had a chance to look into this yet, but I will start today, and hopefully fix the issues:)

vikrosj:
The DataFrame is now the correct length and order, and Claude ran the test again so the test result is updated also. This included some changes from a [previous PR](https://github.com/abrignoni/iLEAPP/pull/1981) as well, which reordered the data_headers and data_list as well. Hopefully things look good now! Glad to hear that you also found Destination Caller ID a good column to surface! :)

abrignoni:
Thanks Viktoria, the destination caller ID is a good column to have.

I pushed a commit to the branch that rebuilds it on top of main, so Sender, Chat Name and Chat Participants stay in the query and the row indices line up. Baseline regenerated from the josh_ios15_ffs case: 48 rows, with Destination Caller ID on 44 of them and on all 15 sent rows.

The conversation view now takes its sent message label from it, since it holds the bare number where Account carries the `P:` prefix.

PR #1985, opened 2026-08-18
Add the remaining BDC stream parsers to batteryBDC

### LINKED ISSUE #1870: Additional BatteryBDC modules
The current BatteryBDC artifact only catches the BDC_SBC_*.csv, however, SBC is only one category of BDC. Given how these are in nearly every extraction and sysdiagnose, it should be pretty easy to implement. 

/var/db/Battery/BDC/
├── BDC.plist                                     ← binary plist (state/metadata)
├── BDC_Once_version<V>_<timestamp>.csv
├── BDC_Daily_version<V>_<timestamp>.csv
├── BDC_Weekly_version<V>_<timestamp>.csv
├── BDC_SBC_version<V>_<timestamp>.csv
├── BDC_OBC_version<V>_<timestamp>.csv
├── BDC_SmartCharging_version<V>_<timestamp>.csv
├── BDC_CPMSRC_version<V>_<timestamp>.csv
├── BDC_AMA_version<V>_<timestamp>.csv            
└── BDC_Timestamps_version1.0.csv

Additionally, */logs/BatteryBDC/ also needs to be included rather than just */var/db in order to have it process sysdiagnoses.

Automated reverse engineering via Fable gives good reference points, although via test data, more reverse engineering is suggested, especially for older versions or the AMA logs:

# Apple Battery Data Collection (BDC) Format Documentation

**Location:** `/var/db/Battery/BDC/`
**Platform:** iOS / iPadOS (also present under battery diagnostics in sysdiagnose archives)
**Producer:** Battery health daemon (`batterydataupdated` / PowerlogCore battery subsystem)
**Purpose:** Long-term battery telemetry — health tracking, charging behavior, gas-gauge diagnostics

> ⚠️ Fields marked *(inferred)* are derived from observed data and reverse engineering, not official Apple documentation.

---

## 1. Directory Layout & File Naming

```
/var/db/Battery/BDC/
├── BDC.plist                                     ← binary plist (state/metadata)
├── BDC_Once_version<V>_<timestamp>.csv
├── BDC_Daily_version<V>_<timestamp>.csv
├── BDC_Weekly_version<V>_<timestamp>.csv
├── BDC_SBC_version<V>_<timestamp>.csv
├── BDC_OBC_version<V>_<timestamp>.csv
├── BDC_SmartCharging_version<V>_<timestamp>.csv
├── BDC_CPMSRC_version<V>_<timestamp>.csv
├── BDC_AMA_version<V>_<timestamp>.csv            ← referenced in plist
└── BDC_Timestamps_version1.0.csv
```

### Filename convention

```
BDC_<StreamName>_version<MAJOR.MINOR>_<YYYY-MM-DD_HH_MM_SS>.csv
```

| Component | Meaning |
|---|---|
| `StreamName` | Logical telemetry stream (see §4) |
| `version` | **Schema version** of the CSV, not software version. Columns are added over time; a version bump signals a schema change. |
| Timestamp | Creation time of the file = time of the **first row** (UTC). Colons in the on-disk name (`13:16:31`) are frequently converted to underscores (`13_16_31`) during export/extraction. |

A new file is created when:
1. The schema version changes (OS update introduces new columns),
2. The battery is replaced (new `Once` file, new `BatterySerialNumber`),
3. File rotation/size limits are reached.

**Consequence:** multiple files of the same stream with different versions coexist. Older-version files are *not rewritten* — they retain the older, narrower schema (see §6, version matrix).

---

## 2. Metadata plist (`BDC.plist`)

A binary plist (`bplist00`) tracking logger state. Observed keys:

| Key | Type | Description |
|---|---|---|
| `DataFormatVersion` | string | Current schema generation, e.g. `"2.0"` |
| `BatterySerialNumber` | string | Battery pack serial, e.g. `FG921471F3218FMA4#ASD`. Changes on battery replacement. |
| `BDC_<Stream>_LastUsedFileName` | string | Absolute path of the CSV currently being appended, e.g. `/var/db/Battery/BDC/BDC_Daily_version2.0_2026-02-26_13:16:31.csv` |
| `BDC_<Stream>_LastLoggedTimeStamp` | date/real | CFAbsoluteTime of the most recent row written for that stream |
| `ReferenceWalltime` | real | CFAbsoluteTime anchor for the RTC↔walltime mapping |
| `LastKnownWalltime` | real | Most recent trusted wall-clock time |
| `ReferenceRTCTicks` | real | RTC tick count at `ReferenceWalltime` |
| `LastKnownRTCTicks` | real | Most recent RTC tick count |
| `RTCTicksDelta` | real | Drift/offset between RTC and wall clock |
| `LastConnectedState` | int/bool | Whether charger was attached at last update |
| `NeedSystemTimeUpdate` | bool | Set when RTC-based rows were written before system time was valid and require correction |

**Timestamp encoding:** plist dates/reals are **CFAbsoluteTime** — seconds since `2001-01-01 00:00:00 UTC` (the `3A C7 ...` byte sequences in the raw dump are big-endian IEEE-754 doubles).

---

## 3. Time Reconstruction — `BDC_Timestamps_version1.0.csv`

The battery gas gauge and early-boot logging run off the **RTC tick counter**, which starts before the system clock is set (notably after battery replacement or full drain). This file provides the mapping to convert tick-stamped events to wall time.

```csv
reference_rtc_ticks,reference_system_time,current_rtc_ticks,set_system_time
54095.520492,2022-05-10 16:30:06 +0000,10.617141,2022-05-10 16:33:16 +0000
```

| Column | Type | Description |
|---|---|---|
| `reference_rtc_ticks` | float (seconds) | RTC counter value at the reference anchor |
| `reference_system_time` | datetime +0000 | Wall clock at the anchor |
| `current_rtc_ticks` | float (seconds) | RTC counter at the moment the system time was (re)set. Note this can be *smaller* than the reference after an RTC reset (battery removal). |
| `set_system_time` | datetime +0000 | The wall time the system clock was set to |

**Reconstruction formula** *(inferred)*:
```
wall_time(event) = reference_system_time + (event_rtc_ticks − reference_rtc_ticks)
```
Each row records a clock-set event; multiple rows allow correcting drift across the log history.

**Forensic value:** reveals device power-loss events, battery pulls, and user/network time changes.

---

## 4. Data Streams

All streams share these conventions:

- **`TimeStamp`** — `YYYY-MM-DD HH:MM:SS`, **UTC**, first column of every stream.
- Empty fields (`,,`) mean *value unavailable at sample time* (e.g., adapter fields while unplugged).
- Numeric conventions: voltages in **mV**, currents in **mA** (negative = discharge), capacities in **mAh**, temperatures in **centi-°C** (`2450` = 24.50 °C).

---

### 4.1 `BDC_Once` — Static Battery Identity

Written **once per battery** (at first boot with a new pack, or after data reset).

| Column | Version | Type | Description |
|---|---|---|---|
| `TimeStamp` | 1.2+ | datetime | Row creation |
| `ChemID` | 1.2+ | int | Battery chemistry ID reported by the gas gauge |
| `AlgoChemID` | 1.2+ | int | Chemistry ID used by the fuel-gauge algorithm (may differ from physical `ChemID` after re-golden updates) |
| `EEEE` | 1.2+ | string | 4-char battery/cell factory code (matches the `EEEE` field in Apple serial conventions), e.g. `18FM`, `M6YR` |
| `YWW` | 1.2+ | int | Manufacture date code: year digit + week, e.g. `215` = year ...2, week 15 *(inferred)* |
| `DesignCapacity` | 1.2+ | int (mAh) | Rated design capacity |
| `GasGaugeFirmwareVersion` | **1.7+** | int | Fuel-gauge (TI Impedance Track family) firmware build, e.g. `109060052` |

---

### 4.2 `BDC_Daily` — Battery Health Snapshot

Cadence: **~2 rows/day** (≈12 h apart; scheduling drifts with device sleep).

| Column | Type | Description |
|---|---|---|
| `WeightedRa` | int | Weighted internal-resistance figure from the gas gauge Ra table *(inferred: scaled mΩ)* |
| `Qmax0` | int (mAh) | Learned maximum chemical capacity of cell 0 |
| `CycleCount` | int | Charge cycle count |
| `NominalChargeCapacity` | int (mAh) | Current full-charge capacity (basis for the user-visible "Maximum Capacity %": `NCC / DesignCapacity`) |
| `TimeAtHighSoc` | hex blob | Little-endian array of **uint32 counters** — cumulative time (hours) spent in state-of-charge bands *(inferred)*. Blob length varies by gauge firmware: observed 40-byte (10×uint32) and 16-byte (4×uint32) variants. Example: `2c030000` → `0x032c` = 812. |
| `ChargingVoltage` | int (mV) | Current charge-voltage target. Values below the chemistry max (e.g. 4239 vs 4495) indicate active longevity derating |
| `BHServiceFlags` | int (bitmask) | Battery-health service flags. `3` = normal baseline observed; `34078723` (= `0x2080003`) appears on first row of a file *(session-init bits, inferred)* |
| `BHCalibrationFlags` | int (bitmask) | Battery-health recalibration state. `0` = none; `1120` observed on a degraded pack undergoing/flagged recalibration *(inferred)* |

---

### 4.3 `BDC_Weekly` — Gauge Resistance Table

Cadence: **1 row / 7 days**.

| Column | Type | Description |
|---|---|---|
| `RaTableRaw0` | hex blob (32 bytes) | Raw **Ra table** of cell 0: 16 × big-endian uint16 values — the Impedance Track per-DOD-band resistance rows (`0x0099` = 153, …). The final words include scale/status entries *(inferred)* |
| `TotalOperatingTime` | int (hours) | Cumulative battery operating time since manufacture |
| `GasGaugeFirmwareVersion` | int | As in `Once`; logged weekly to catch gauge firmware updates |

---

### 4.4 `BDC_SBC` — State/Snapshot of Battery Conditions

High-frequency stream: **~every 5 min while awake**, plus event-driven rows (charge start/stop). The richest stream, and the one with the most schema churn.

| Column | Version | Type | Description |
|---|---|---|---|
| `PresentDOD0` | 2.0+ | int | Present depth-of-discharge of cell 0 (gauge internal units; rises as battery drains) |
| `CurrentCapacity` | 2.0+ | int (%) | UI-facing state of charge |
| `IsCharging` | 2.0+ | bool | 1 while actively charging |
| `Temperature` | 2.0+ | centi-°C | Battery temperature (`2450` = 24.5 °C) |
| `Amperage` | 2.0+ | mA | Filtered/average current (negative = discharge) |
| `InstantAmperage` | 2.0+ | mA | Instantaneous current |
| `Voltage` | 2.0+ | mV | Battery terminal voltage |
| `StateOfCharge` | 2.0+ | int (%) | Gauge-reported SoC (may differ slightly from `CurrentCapacity` due to UI smoothing) |
| `ChargeAccum` | 2.0+ | int (mAh) | Coulomb counter accumulator (signed; monotonic between resets) |
| `ResScale` | 2.0+ | int | Gauge resistance scale factor |
| `RSS` | 2.0+ | int | Resistance scale/status value from gauge *(inferred)* |
| `ITMiscStatus` | 2.0+ | int (bitmask) | Impedance-Track gauge misc status register |
| `AccumulatedSystemLoad` | **2.6+** | int | Cumulative system energy drawn since boot/reset *(inferred: mW·s or similar)* |
| `AccumulatedSystemPowerIn` | **2.6+** | int | Cumulative energy delivered from external power; `0` while on battery |
| `VirtualTemperature` | **2.6+** | centi-°C | Modeled/virtual cell temperature used for thermal-aware charging |
| `SystemVoltageIn` | **2.8+** | mV | Input voltage from adapter (e.g. `8476` = 9 V USB-PD contract sagging under load); `0` when unplugged |
| `SystemCurrentIn` | **2.8+** | mA | Input current from adapter; `0` when unplugged |

---

### 4.5 `BDC_OBC` — On-charger / External Power Events

**Event-driven**: one row per attach/detach/charge-state transition.

| Column | Type | Description |
|---|---|---|
| `FamilyCode` | int (signed 32-bit) | IOPS power-adapter family code, stored as signed decimal. Convert: `value + 2³²` → e.g. `-536854524` = `0xE0004004`, `-536854525` = `0xE0004003`, `-536854521` = `0xE0004007`. The `0xE0004xxx` range covers USB-C/PD adapter families *(inferred from IOKit `kIOPSFamilyCode*`)*. Empty when no adapter identified. |
| `ExternalConnected` | bool | System-level "external power present" |
| `AppleRawExternalConnected` | bool | Raw hardware attach state (pre-debounce/pre-policy) |
| `ChargingOverride` | int | Non-empty when a policy override is active (0 = none). Empty in some versions when not applicable |
| `NotChargingReason` | int (bitmask) | `0` = charging normally. Observed: `1` , `128`, `1024`, `1152` (=1024+128), `1153`. Bits encode inhibit reasons (charge-limit hold, thermals, Optimized Charging, full) *(inferred)* |
| `VacVoltageLimit` | int (mV) | Adapter input-voltage limit / VBUS clamp (e.g. `4410`, `4340` — tracks battery charge-voltage ceiling) |

> **Note:** OBC schema is column-identical across observed versions 2.0 / 2.6 / 2.8 — the version number tracks the overall `DataFormatVersion`, not per-stream changes.

---

### 4.6 `BDC_SmartCharging` — Optimized Battery Charging (OBC/80 % limit)

**Event-driven**: rows on charging-policy decisions.

| Column | Version | Type | Description |
|---|---|---|---|
| `ChargingState` | 2.0+ | int | `1` = charging allowed, `0` = held/paused by policy |
| `InflowState` | 2.0+ | int | `1` = power inflow enabled (inflow can be cut while attached) |
| `ChargeLimit` | 2.0+ | int (%) | Active charge ceiling (`100`, or `80` when a limit is engaged) |
| `CheckPoint` | 2.0+ | int | State-machine checkpoint ID (`0`, `2`, `5`, `6`, `30` observed — e.g. hold-engaged / resume / target-reached) *(inferred)* |
| `DecisionMaker` | **2.6+** | int | Which subsystem made the decision: `-1` = none/default, `0` = Optimized Battery Charging engine *(inferred)* |
| `ModeOfOperation` | 2.0+ | int | Charging mode: `0` = normal, `1` = OBC hold, `6` = session start/evaluation *(inferred)* |

A typical overnight pattern in v2.0 data:
```
0,1,100,2,1   ← evening: charging paused at 80 % (hold)
1,1,100,5,0   ← pre-wake: resume to 100 %
```

---

### 4.7 `BDC_CPMSRC` — Charge Power Management: Source Impedance Model

Cadence: **daily**. Parameters of the fitted battery **RC equivalent-circuit model** (series R plus four parallel RC branches) used by the charging/power-management system.

| Column | Type | Description |
|---|---|---|
| `ImpedanceR0PlusRtrace (mΩ)` | float | Series resistance: cell R0 + board trace resistance |
| `ImpedanceR1..R4 (mΩ)` | float | Resistances of RC branches 1–4 (fast → slow time constants) |
| `ImpedanceRCFreq1..4 (Hz)` | float | Corner frequencies of each RC branch (e.g. ~1.1 kHz, ~35 Hz, ~0.8 Hz, ~0.03 Hz) |

> Encoding note: header units appear as `mÎ©` when the file's UTF-8 `mΩ` is decoded as Latin-1. Always read these CSVs as **UTF-8**.

---

### 4.8 `BDC_AMA`

Referenced in the plist (`BDC_AMA_LastLoggedTimeStamp`) but no sample present in this dataset. Schema undocumented here.

---

## 5. Hex Blob Reference

| Blob | Layout | Notes |
|---|---|---|
| `TimeAtHighSoc` (Daily) | Array of **little-endian uint32** | Cumulative time counters per SoC band. Length varies with gauge firmware: 40 B (10 counters) or 16 B (4 counters) observed. Values are monotonically non-decreasing across rows. |
| `RaTableRaw0` (Weekly) | 16 × **big-endian uint16** (32 B) | Per-DOD resistance rows from the Impedance Track gauge; trailing words carry scale/flag data. Rising values over months = cell aging. |

---

## 6. Schema Version Matrix

| Stream | v1.2 | v1.7 | v2.0 | v2.6 | v2.8 |
|---|---|---|---|---|---|
| Once | 6 cols | + `GasGaugeFirmwareVersion` | — | — | — |
| Daily | | | 9 cols (stable) | ″ | ″ |
| Weekly | | | 4 cols (stable) | ″ | ″ |
| SBC | | | 13 cols | + `AccumulatedSystemLoad`, `AccumulatedSystemPowerIn`, `VirtualTemperature` (16) | + `SystemVoltageIn`, `SystemCurrentIn` (18) |
| OBC | | | 7 cols (stable) | ″ | ″ |
| SmartCharging | | | 6 cols | + `DecisionMaker` (7) | ″ |
| CPMSRC | 10 cols | — | — | — | — |
| Timestamps | v1.0 only, 4 cols | | | | |

**Parsing rule:** old-version files are strict column subsets (where changed). A robust parser should key on the header row, not the version number, and null-fill missing columns when normalizing into a unified table.

---

## 7. Parsing & Forensic Notes

1. **All CSV timestamps are UTC.** Local activity patterns (overnight charging in `SmartCharging`, wake gaps in `SBC`) can be used to infer device timezone.
2. **Retention** spans years — observed data ranges from 2021 through 2026 across file generations. BDC survives OS updates; it resets with battery replacement (watch for `BatterySerialNumber` change + new `Once` file + RTC reset in `Timestamps`).
3. **Multiple devices/batteries can appear in one collection** (e.g., differing `GasGaugeFirmwareVersion`, `ChemID`, capacity ranges). Group files by the `Once` identity before analysis.
4. **Health computation:** `MaximumCapacity% ≈ NominalChargeCapacity / DesignCapacity × 100` (Daily × Once join).
5. **Charging-session reconstruction:** join `OBC` (attach/detach) + `SBC` (`IsCharging`, `SystemVoltageIn/CurrentIn`) + `SmartCharging` (policy holds) on `TimeStamp`.
6. **Clock-tamper detection:** `BDC_Timestamps` rows plus `NeedSystemTimeUpdate` in the plist expose every system-time set event relative to the monotonic RTC.
7. **Encoding:** UTF-8; beware mojibake in `CPMSRC` headers. Line endings `\n`. No quoting observed; hex blobs are unquoted tokens.

ChrisJr404:
Closes #1870. batteryBDC only caught `BDC_SBC_*.csv`, so this adds the other BDC streams that sit in the same folder and show up in nearly every full filesystem and sysdiagnose.

New artifacts, each with its own glob and output section:

- Once (static battery identity, ChemID/EEEE/DesignCapacity)
- Daily (health snapshot, cycle count, nominal charge capacity)
- Weekly (Ra table, total operating time)
- OBC (charger attach/detach events; FamilyCode also decoded to its `0xE0004xxx` hex form)
- SmartCharging (Optimized Battery Charging policy, charge-limit holds)
- CPMSRC (RC impedance model)
- Timestamps (RTC-to-wall-clock set events, useful for clock-tamper checks)

I also added the sysdiagnose `*/BatteryBDC/` path next to the existing `*/Battery/BDC/` on every stream including SBC, since the issue notes sysdiagnoses live under `logs/BatteryBDC/`. SBC's parsing is otherwise unchanged.

One thing worth calling out: columns are resolved by header name, not fixed index. The schema grows across versions and SmartCharging inserts `DecisionMaker` ahead of `ModeOfOperation` at v2.6, so positional parsing would silently shift that column on older files. Name lookup null-fills anything a narrower-schema file doesn't have.

Left AMA out on purpose. There's no AMA sample in the corpus and the schema isn't documented anywhere I could find, so I'd only be guessing at columns.

Testing:
- Globs run through `fnmatch.fnmatchcase` against the recorded path lists in `admin/data/filepath-lists/` (with the synthetic `root/` prefix): Once 5, Daily 21, Weekly 17, OBC 24, SmartCharging 26, Timestamps 1 across the four images. The new `*/BatteryBDC/` variant adds zero overlap with the var/db pattern, so no double-counting.
- Built synthetic CSVs matching the documented schema for each stream (including a v2.0 and a v2.6 SmartCharging file) and ran each function's `.__wrapped__` with a mock context. Row widths match headers, FamilyCode `-536854524` decodes to `0xE0004004`, and SmartCharging reads ModeOfOperation correctly on both versions.
- Full `ileapp.py -t fs` run over those samples: all seven artifacts parse, emit HTML/TSV/timeline/LAVA, and the whole 924-artifact set loads with no duplicate-name error.

CPMSRC isn't in any of the local test images, so its parser is written off the documented column names and hasn't been checked against a real file yet.


abrignoni:
Test data is now included for every changed artifact module. Thank you!

PR #2001, opened 2026-08-19
Make the Pinterest cache container check fail closed

abrignoni:
Makes the Pinterest cached image and model container check fail closed.

`_cache_files` applied its container test only when a Pinterest container had already been identified, so an extraction holding none kept every cache file instead of none. PINRemoteImage is Pinterest's own open source image library and other applications embed it, so those files belong to whoever wrote them.

- No Pinterest container identified now means no rows, with a log line saying so
- `pinterestCachedModelRequests` declares the container marker patterns in its own paths, so the check does not depend on artifact run order

This is the behaviour the artifact notes already describe.


abrignoni:
Superseded by #2002, which fixes both fail-open sites in pinterest.py rather than one, and additionally declares the container markers on Cached Model Requests. Verified equivalent-or-better by both authors against the registered corpora. Closing in favour of it.

Issue #1106, opened 2025-03-30
Cannot install Ileaps in Autopsy


Maselia464:
I cannot install Ileaps in autopsy there is no option or documentation for it ?

Can you update it ?

stark4n6:
@markmckinnon do you have documentation for Autopsy to work with the LEAPPs? We may also have to update the executables at some point too

Johann-PLW:
@Maselia464 

Both iLEAPP and ALEAPP are already installed in Autopsy, but the default versions are rather old.
If you want to use the latest version of iLEAPP, download the [binary for Windows](https://github.com/abrignoni/iLEAPP/releases) **ileapp-v2.x.x-Windows.zip** and unzip it.
Then copy the new unzipped **ileapp.exe** file into Autopsy installation folder (i.e.: C:\Program Files\Autopsy-4.22.0\autopsy\iLeapp) to replace the older one.
Then, in Autopsy, you can run the iLEAPP ingest module to decode iOS artifacts.

But Autopsy's GUI only displays a few of the results of iLEAPP's processing. The complete iLEAPP's report is available in the `report`folder in your case directory.

JamesHabben:
closing with instructions to address issue. 

@Maselia464 feel free to reopen if you need further help

PR #2027, opened 2026-08-23
Keep case-variant evidence files when the report volume folds case

### LINKED ISSUE #1948: Seeker overwrites evidence files whose paths differ only in case, so one file is silently lost from the report
## What happens

`FileSeekerDir.search()` builds the report-data destination from the source path and copies
into it with no check for an existing, differently-sourced occupant:

- [`search_files.py:345`](https://github.com/abrignoni/iLEAPP/blob/1b55104f69e91e3f569392cb8f3dcb9649fd3c90/scripts/search_files.py#L345) (destination built from the source path)
- [`search_files.py:348`](https://github.com/abrignoni/iLEAPP/blob/1b55104f69e91e3f569392cb8f3dcb9649fd3c90/scripts/search_files.py#L348) (the guard is keyed by SOURCE path)
- [`search_files.py:354`](https://github.com/abrignoni/iLEAPP/blob/1b55104f69e91e3f569392cb8f3dcb9649fd3c90/scripts/search_files.py#L354) (copy2 overwrites whatever is there)

Two evidence files whose paths differ only in case are two distinct sources, so both pass the
`if item not in self.copied` guard, and both resolve to the **same** destination on a
case-insensitive output volume. The second `copy2` destroys the first. `FileSeekerZip` has the
same shape at [`search_files.py:809`](https://github.com/abrignoni/iLEAPP/blob/1b55104f69e91e3f569392cb8f3dcb9649fd3c90/scripts/search_files.py#L809) (zip extract), and `FileSeekerTar` at its own write.

`FileSeekerItunes` has the same shape at its own `copy2` call, and `FileSeekerFile` at its single-file copy.

This is not hypothetical. An iOS extraction can contain both `com.apple.MobileSMS.plist` and
`com.apple.mobileSMS.plist` in one directory as two different files with different contents:
**6 of 20** local iOS corpora carry exactly that pair, from iOS 14.3 through 26.5.2, so the
acquired filesystem clearly preserved both. Windows and stock macOS report volumes are
case-insensitive, so that pair collides on extraction for most users.

## Why it matters for a forensic report

Two consequences, both confirmed by running the real seekers:

1. **The losing file's bytes are gone.** Only one file survives under `data/`, so the report
   preserves one of the two and silently drops the other. Which one survives depends on
   directory or archive member order, which the acquisition tool decides.

2. **A row's source path can point at a file that does not contain what the row reports.**
   An artifact that reads each file before the next copy overwrites it gets the correct
   values, but the preserved copy at the path shown in the report then holds the *other*
   file's contents. An examiner who opens the cited path sees something that does not support
   the row. That is a defensible-evidence problem, not a cosmetic one.

There is a third, quieter effect: `self.file_infos` ends up with two entries whose keys differ
only in case but which `stat()` to the same inode, carrying two different claimed source paths
and two different timestamps. Those timestamps reach examiner-facing output.

## Reproducing it

Any two files differing only in case will do. On macOS, note the default volume is
case-insensitive, so stage the *input* on a case-sensitive one or read from a zip:

```python
import zipfile, plistlib
PREF = "private/var/mobile/Library/Preferences/"
with zipfile.ZipFile("ev.zip", "w") as zf:
    zf.writestr(PREF + "com.apple.MobileSMS.plist", plistlib.dumps({"SSKeepMessages": 365}))
    zf.writestr(PREF + "com.apple.mobileSMS.plist", plistlib.dumps({"Other": True}))
```

Point a seeker at it with the report folder on a case-insensitive volume, search for both
names, then list the data folder: one file, and both returned paths `stat()` to the same
inode.

## Suggested direction, for whoever owns the core

This is core code and the LEAPPs treat core changes as opt-in, so I have not touched it. The
shape of a fix, for discussion rather than as a proposal to merge:

- Probe the data folder's real case behaviour once per seeker at construction (create
  `Aa`/`aA` and see whether two files result) rather than inferring it from
  `os.path.normcase`, which reports the *platform's* convention and not the *volume's*.
- When a destination is already claimed by a different source, disambiguate it rather than
  overwriting, and record the mapping so the report's source path stays truthful.
- Whatever is chosen wants to land in all five cores together, since the code is duplicated.

## Related

An artifact-level workaround exists for the one case found so far: read each match before
searching for the next, so a file is read before the next copy can overwrite it. That is what
https://github.com/abrignoni/iLEAPP/pull/1946 does for `messageRetention` (merged). It fixes
the reported values; it does **not** fix the preserved copies, which is why this issue is
still worth having.

A separate open PR, https://github.com/abrignoni/iLEAPP/pull/1947, adds a CI guard that stops
an artifact *declaring* two `paths` patterns differing only in case. That is a different
problem with the same root cause, and the guard does nothing about the seeker.

## Same issue in the sibling cores

`scripts/search_files.py` is duplicated across the five extractors and every copy has
this shape, so a fix wants levelling. Filed separately so each repo's maintainers see it:

- ALEAPP: https://github.com/abrignoni/ALEAPP/issues/1106
- DLEAPP: https://github.com/abrignoni/DLEAPP/issues/74
- RLEAPP: https://github.com/abrignoni/RLEAPP/issues/416
- VLEAPP: https://github.com/abrignoni/VLEAPP/issues/131

forest-savage1234:
# Keep case-variant evidence files when the report volume folds case

Same dest-guard as the sibling LEAPP seekers. Seeker only. Not an artifact. Please land this with the other three so the seekers stay level.

- [iLEAPP #1948](https://github.com/abrignoni/iLEAPP/issues/1948) / [PR #2027](https://github.com/abrignoni/iLEAPP/pull/2027) (this PR)
- [ALEAPP #1106](https://github.com/abrignoni/ALEAPP/issues/1106) / [PR #1199](https://github.com/abrignoni/ALEAPP/pull/1199)
- [VLEAPP #131](https://github.com/abrignoni/VLEAPP/issues/131) / [PR #140](https://github.com/abrignoni/VLEAPP/pull/140)
- [RLEAPP #416](https://github.com/abrignoni/RLEAPP/issues/416) / [PR #430](https://github.com/abrignoni/RLEAPP/pull/430)

Closes abrignoni/iLEAPP#1948.

## Summary

`self.copied` is keyed by the source path. The report destination is built from that path. On a case-insensitive report volume, `com.apple.MobileSMS.plist` and `com.apple.mobileSMS.plist` are two sources and one destination. The second write silently destroys the first.

This PR does only the seeker dest-guard named on the issue:

1. Probe the data folder's real case behaviour once per seeker (create `Aa` / `aA`; do not trust `os.path.normcase`).
2. When a destination is already claimed by a different source, write to `name~caseN.ext` instead of overwriting.
3. Return and record that path so `file_infos` and the artifact's source path stay truthful.

## Validation

Earlier local check on a case-sensitive ext4 volume (dest names folded to stand in for Windows / stock macOS): two sources plus one folded dest left one file. The dest-guard kept both; the second dest is `*~case2.plist`. The volume probe left no leftover probe names. This is not an artifact change, so the artifact report checkers were not run.

## Files

1. `scripts/search_files.py` (seeker dest-guard only)

Touched write sites: FileSeekerDir, FileSeekerItunes, FileSeekerTar, FileSeekerZip, FileSeekerFile.

## Out of scope

- Artifact parsers
- iLEAPP #1947 path-tuple CI guard
- DLEAPP #74
- Changing `os.path.normcase` matching of search patterns
- RLEAPP #194 FileSeekerItunes port

## How to review

On a case-insensitive report volume, search a zip that holds both `com.apple.MobileSMS.plist` and `com.apple.mobileSMS.plist`. `data/` should keep two files (the second named `*~case2.plist`). Case-sensitive volumes should be unchanged: the probe sees two files and never mints `~caseN`.

I am a volunteer. Thank you for the time. I am trying to become more useful on this work, so I welcome a critical look. If this is the wrong cut, or you want me to stand down, say so and I will recut from notes.

### DIFF
--- a/admin/test/scripts/test_seeker_case_collision.py
+++ b/admin/test/scripts/test_seeker_case_collision.py
@@ -1,0 +1,293 @@
+"""Pin the seeker's dest-guard for case-variant evidence files (issue #1948).
+
+An iOS extraction can hold com.apple.MobileSMS.plist and com.apple.mobileSMS.plist
+in one directory as two different files. On a case-insensitive report volume both
+sources fold to one destination under data/, and before the guard the second copy
+silently destroyed the first, so the preserved file at a cited path could hold the
+other file's bytes.
+
+The guard writes the colliding copy to name~case-<tag>.ext instead, where the tag
+is derived from the evidence-relative source path. These tests pin three
+properties: both byte streams survive, the alternate name is a pure function of
+the source (stable across re-searches, seeker instances and force=True), and
+directory members never mint a tagged twin.
+
+The claims logic folds keys only when the data folder's volume folds case, which
+CI runners may not. Tests that need folding force the seeker's probed flag, which
+exercises the identical code path on any filesystem.
+"""
+import hashlib
+import os
+import pathlib
+import shutil
+import sys
+import tempfile
+import unittest
+import zipfile
+from functools import lru_cache
+
+REPO_ROOT = pathlib.Path(__file__).resolve().parents[3]
+sys.path.insert(0, str(REPO_ROOT))
+
+from scripts.search_files import (  # pylint: disable=wrong-import-position
+    FileSeekerZip, _case_variant_digest, _disambiguated_data_path,
+    _probe_volume_case_insensitive)
+import scripts.search_files  # pylint: disable=wrong-import-position
+
+PREF = 'private/var/mobile/Library/Preferences/'
+UPPER = PREF + 'com.apple.MobileSMS.plist'
+LOWER = PREF + 'com.apple.mobileSMS.plist'
+UPPER_CONTENT = b'upper spelling bytes'
+LOWER_CONTENT = b'lower spelling bytes'
+PATTERN = '*/mobile/Library/Preferences/com.apple.[Mm]obileSMS.plist'
+
+
+def _expected_tag(member):
+    return hashlib.sha256(member.encode('utf-8')).hexdigest()[:8]
+
+
+def _tree_files(base):
+    found = {}
+    for root, _dirs, files in os.walk(base):
+        for name in files:
+            full = os.path.join(root, name)
+            with open(full, 'rb') as fin:
+                found[os.path.relpath(full, base).replace(os.sep, '/')] = fin.read()
+    return found
+
+
+class TestDisambiguatorUnit(unittest.TestCase):
+    """The claims logic, exercised directly with both volume behaviours."""
+
+    def setUp(self):
+        self.tmpdir = tempfile.mkdtemp()
+
+    def tearDown(self):
+        shutil.rmtree(self.tmpdir, ignore_errors=True)
+
+    def _dest(self, name):
+        return os.path.join(self.tmpdir, name)
+
+    def test_case_sensitive_volume_keeps_both_plain_names(self):
+        claims = {}
+        first = _disambiguated_data_path(self._dest('A.plist'), 'src/A.plist',
+                                         claims, False)
+        second = _disambiguated_data_path(self._dest('a.plist'), 'src/a.plist',
+                                          claims, False)
+        self.assertEqual(os.path.basename(first), 'A.plist')
+        self.assertEqual(os.path.basename(second), 'a.plist')
+
+    def test_folding_volume_tags_the_second_source(self):
+        claims = {}
+        first = _disambiguated_data_path(self._dest('A.plist'), 'src/A.plist',
+                                         claims, True)
+        second = _disambiguated_data_path(self._dest('a.plist'), 'src/a.plist',
+                                          claims, True)
+        self.assertEqual(os.path.basename(first), 'A.plist')
+        self.assertEqual(os.path.basename(second),
+                         f'a~case-{_expected_tag("src/a.plist")}.plist')
+
+    def test_same_source_returns_the_same_path_every_time(self):
+        claims = {}
+        _disambiguated_data_path(self._dest('A.plist'), 'src/A.plist', claims, True)
+        paths = {_disambiguated_data_path(self._dest('a.plist'), 'src/a.plist',
+                                          claims, True)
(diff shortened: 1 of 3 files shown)


forest-savage1234:
Thank you for the time on this review. This only keeps case-variant evidence files when the report volume folds case.

abrignoni:
Thanks for this one. The direction was right, so rather than a recut round trip I pushed two commits on top of your branch.

What testing it turned up:

1. Repeated searches with force=True minted a new ~caseN copy each time, because the claims table only remembered the base destination. Runs that re-search a pattern grew ~case2, ~case3, ~case4 copies of the same source.
2. messageRetention, the one artifact that feeds a returned basename back into a search pattern, silently lost a row: a tagged basename is not an archive member name, so its per-spelling re-search found nothing. It was also the artifact written to work around this exact bug, so the two had to move together.
3. A zip that stores a directory entry after its contents produced a spurious ~case2 directory, since the parent already existed on disk unclaimed.

The fixes: the alternate name is now derived from a hash of the evidence-relative source path (name~case-<tag>.ext), so the same source maps to the same destination on every search, every run and every machine, with no counter state to get wrong. Directory members skip the guard. messageRetention reads every preserved copy directly and labels each row with the source spelling the seeker recorded, which is simpler than the read-before-the-next-copy dance it needed before.

Validation: messageRetention run through the real ileapp.py on all six registered iOS images that carry both spellings (zip and tar inputs) plus a single-spelling control, main vs fixed, report folder on a case-insensitive volume. Row counts and values identical to main everywhere, both files preserved with distinct hashes on every pair image, and the control run byte-identical. The new suite in admin/test/scripts fails against the old seeker and covers both volume behaviours, name determinism across instances and forced re-searches, directory members, and a simulated Windows normcase. A full-suite before/after run comparison is going now ahead of the merge.

The same fixes are needed on the ALEAPP, VLEAPP and RLEAPP PRs; I will level them there.

PR #2052, opened 2026-08-25
cashApp.py Update

charpy4n6:

- [ ] Ran the tool against a real extraction and confirmed the row counts, not just that it imports.
- [ ] Checked it against a second app data directory where the platform provides one (a second user, account, or container). `--compare <multi-container report>` reads the scaling for you: it should be exactly double.


abrignoni:
Thanks for taking this on. The protobuf rewrite is the right direction and the decoding itself is solid. I ran the branch head (`ad8c4e3`) against both images named in `sample_data` and got 1 row on each, which matches what the artifact claims, with no database errors in either log. Amounts, timestamps and the customer fields all land where they should. On the second store I also checked the join arithmetic: `ZPAYMENT` holds exactly one row against 65 in `ZCUSTOMER`, so the inner join is not dropping anything.

Four things need a fix before this can go in.

**1. The `source_path` return will fail lint.** The module returns `'see Source File for more info'` as the third element. #2049 merged a checker for exactly that today, and it flags this line:

```
cashApp.py:222  get_cashApp()  'see Source File for more info'
```

I confirmed it downstream too. The LAVA manifest records that sentence where the artifact's path should be, so the "located at" line in the report points at nothing. The previous version built a `source_files` set and joined it, which was right. Bringing that back is all this needs.

**2. The first path pattern does not match anything.** It looks for `CCEntitySync.sqlite`, but the file on both test images is `CCEntitySync-internal.cashappapi.com.sqlite`, so that literal never matches. Checked with `fnmatch` against the real paths:

| pattern | hc_ios18_7 | abe_ios16 |
| --- | --- | --- |
| `*/Environments/Production/Accounts/*/CCEntitySync.sqlite*` | no | no |
| `*/mobile/Containers/Shared/AppGroup/*/CCEntitySync-internal.cashappapi.com.sqlite*` | yes | yes |
| `*/mobile/Containers/Shared/AppGroup/*/CCEntitySync-api.squareup.com.sqlite*` | no | no |

Every row in both runs came from the second pattern. The comment above the tuple says the first pattern covers the AppGroup form, and it works the other way around. It also does not reproduce what Cash App (B) was matching, because that one had a wildcard ahead of `-internal`.

**3. `cashAppB.py` is still in the tree.** The notes say this is merged from Cash App and Cash App (B), but the PR only touches `cashApp.py`, so Cash App (B) still ships and still matches the same database on one of the test images. That gives you the same payment reported twice under two artifact names. Either delete `cashAppB.py` in this PR or drop the merge wording from the notes.

**4. Missing newline at end of file.** Pylint `C0304`, and warnings fail the build.

Two smaller things on the claims.

`check_artifact_output.py` reports 11 empty columns on both images: card brand, suffix, the instrument fields, transaction id, receipt, and the captured, reached customer, paid out and deposited timestamps. Those read as real negatives rather than bugs, since the single payment on each image is still pending and carries no card instrument, so nothing was ever captured or deposited. Worth saying that in the notes, which is also what clears the checker.

The notes say records that do not match the expected protobuf layout fall back to a raw scan. Neither image exercised that path, both blobs decoded cleanly, so that sentence should say the fallback is present but not yet seen in real data.

One more with no action needed, just so you know. `output_types` went from `standard` to `all`, but `all` only adds KML, and `kmlgen` returns right away unless the headers carry Latitude and Longitude. It does nothing here either way.

Good stuff overall. Fix those four and I am happy to take it.

abrignoni:
Thanks, this is the right rewrite. Decoding the blobs instead of slicing them in SQL also fixed two columns that were dumping raw protobuf into the report, which is more than the PR body claimed.

Adding a few fixes on top before this goes in.

The `or` in `payment_info.get('state') or _scan_raw(...)` fires when the key is missing, not when the parse fails, and Cash App writes no top level `state`. So the scan ran on every row and picked up the `state` nested inside `pending_referral_render_data`, which belongs to the referral bonus rather than to the payment. That was landing in the Transaction State column. It now falls back only when the payload is absent, and the column reads blank on both test images.

First path pattern is now `CCEntitySync-*.sqlite*`, which matches the real filenames and restores what Cash App (B) covered for an extraction rooted below the app group container. `cashAppB.py` is removed, so no more double reporting on `hc_ios18_7`.

Smaller ones: timestamps grouped after the lead column, `Units` renamed to say whose preference it records, notes naming the columns that came back empty and why, and the trailing newline. Your `d1a4dcc` had already sorted the `source_path` return.

Checked with 1 row on each test image, and 2 rows on a tree holding one store matched by two patterns plus a second account and a decoy.

You are in the author field. Good stuff, send more.

PR #2086, opened 2026-09-01
Initial Sysdiagnose Parsers

stark4n6:
## What this changes

Adds 10 Initial Sysdiagnose parsers

- [ ] Ran the tool against a real extraction and confirmed the row counts, not just that it imports.
- [ ] Checked it against a second app data directory where the platform provides one (a second user, account, or container). `--compare <multi-container report>` reads the scaling for you: it should be exactly double.


abrignoni:
Heads up, #2102 adds a `get_sysdiagnose_files` helper in ilapfuncs that reads a target file either standalone or from inside a `sysdiagnose_*.tar.gz`. If these modules go through it they will also work on a full file system image or a folder of raw tars, not just an extracted sysdiagnose folder. Worth doing before this one lands.

abrignoni:
Superseded by #2205, which keeps these commits and adds the fixes it needed: the shutdown log regex, the FDR report parsing, the packed sysdiagnose paths for User Settings, notes and sample_data for every artifact, and the lint cleanup. Merged there. Thanks for the parsers, @stark4n6.

PR #2102, opened 2026-09-04
Sysdiagnose Extraction Helper

stark4n6:
## What this changes

Add helper function in ilapfuncs to process an embedded Sysdiagnose tar.gz if added as part of the path regex. The idea is it will hunt them if in a FFS or if you provide a folder of files.

Sample of usage is applied to the current sysdiagnose.py parser for Account Devices.

- [x] Ran the tool against a real extraction and confirmed the row counts, not just that it imports.
- [x] Checked it against a second app data directory where the platform provides one (a second user, account, or container). `--compare <multi-container report>` reads the scaling for you: it should be exactly double.


abrignoni:
Ran it against main on felix23, the two Hickman images and three standalone sysdiagnose sets. Same rows everywhere, the folder-of-files case works, and the in-progress archives get skipped. Good stuff.

Three modules still open the tar by hand: mobileActivationLogs, mobileInstall and systemVersionPlist. Want to move them onto your helper here or in a follow-up? Your call. One note worth adding to the artifact: archives renamed by a vendor (something like `<udid>_sysdiagnose.tar.gz`) will not match.

Mark it ready when you are done and I will merge.

stark4n6:
@abrignoni are there many other vendors that rename them? I don't know of specific ones but we can account for them later as we find them

stark4n6:
@abrignoni I will also work on updating the other parsers to work with this, as I have more in #2086 to make work too

abrignoni:
Merged, thanks Kevin.

Only one I have seen so far is an archive named `<udid>_sysdiagnose.tar.gz`, and I do not know which tool wrote it. Agreed, handle them as they show up.

PR #2154, opened 2026-09-08
Run crunch_artifacts on a worker thread so the GUI stays responsive

OneSixForensics:
## The problem

The GUI called `crunch_artifacts()` directly on the Tk main thread from the Process button handler, so for the whole length of a run the only thing pumping the event loop was `log_text.update()` inside `logfunc` (`scripts/ilapfuncs.py`). Between log lines the window stops answering Windows, which paints "(Not Responding)" over a run that is perfectly healthy. A single slow artifact is enough to trigger it, and an examiner watching a report that looks hung has no way to tell it apart from one that is.

## The change

`crunch_artifacts` runs on a daemon worker thread and reports back over a `queue.Queue` that an `after()` poller drains on the main thread.

- `GuiWindow.message_queue` is set while a run is in flight. `logfunc` and `SetProgressBar` check it and enqueue instead of touching a widget, so the worker never calls into Tk.
- The poller drains the queue every 50 ms and does one `log_text.insert()` per poll rather than one per line, which is also less work than the old path.
- The completion UI moved into `finish_crunch()`, called by the poller on the main thread.
- `initialize_lava()` moved out of `process()` and into the worker. A `sqlite3` connection may only be used by the thread that created it, so LAVA's connection has to be opened on the same thread the artifacts run on. Opening it on the main thread raises `SQLite objects created in a thread can only be used in that same thread` on the first artifact that writes.

The `scripts/ilapfuncs.py` half is identical in all five LEAPP cores. The GUI half keeps each core's own `crunch_artifacts` signature.

## Measured

`logfunc` and `SetProgressBar` are the real ones, driven through the real `message_queue` gate; only the workload is simulated (12 artifacts, one of them working 2.5 s without logging). Worst event-loop stall, and how many times the loop was serviced:

| | before | after |
|---|---|---|
| worst stall | 2.55 s | 0.01 s |
| loop serviced | ~300x | ~3300x |

Windows paints "(Not Responding)" at roughly 5 s, so 2.55 s is the visible symptom's floor rather than its ceiling — a genuinely slow artifact stalls for longer.

## Verified on a real run

Full GUI run against a 19 GB iPhone extraction: 919 of 920 artifacts, 465 s wall, event loop serviced 8,391 times, worst stall 0.31 s, no errors.

The other four LEAPP cores carry the same change on a branch of this name; each was verified against its own real data, since each keeps its own `crunch_artifacts` signature and passing one does not clear the others.

### DIFF
--- a/ileappGUI.py
+++ b/ileappGUI.py
@@ -3,6 +3,9 @@
 import tkinter as tk
 import typing
 import json
+import queue
+import threading
+import traceback
 import ileapp
 import webbrowser
 import base64
@@ -28,6 +31,11 @@ from leapp_functions.app.platform import sanitize_file_name
 from leapp_functions.app.output import default_output_folder_name, validate_output_folder_available
 from scripts.context import Context
 from scripts.lavafuncs import lava_json_name
+
+
+# How often the main thread drains the worker's queue. Short enough that the log still
+# scrolls smoothly, long enough that polling is not itself the load.
+CRUNCH_POLL_MS = 50
 
 
 def allow_output_folder_name_chars(proposed):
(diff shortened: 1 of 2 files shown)


abrignoni:
Thanks @OneSixForensics. The not responding window has bugged folks for a long time and this is a clean fix. Merged in all five cores, with two small follow-ups on top. Good stuff.

PR #2155, opened 2026-09-09
Add Apple SMS preview attachments (ATX in KTX) artifact

charpy4n6:
## What

New artifact `apple_sms_preview_ktx_atx` (category **SMS & iMessage**).

iOS caches preview thumbnails for Messages attachments under
`com.apple.MobileSMS/Previews/Attachments/…`. Some of these files carry a `.ktx`
extension but the container is Apple's **ATX** format (`AAPL\r\n\x1a\n` magic
followed by `HEAD` / `astc` / `LZFS` chunks), not Khronos KTX1. The existing
`apple_atx_images` artifact only matches `**/*.atx`, so these previews were never
surfaced.

## How

Path-scoped sibling of `apple_atx_images`. It reuses
`leapp_functions/parsers/apple_atx.decode_atx_file` **unchanged** — that parser
sniffs the `AAPL` magic, not the file extension, so a `.ktx`-named ATX container
parses and decodes the same way. Columns are identical to `apple_atx_images`
(file timestamps, decoded image, HEAD metadata, payload chunk info, status).

`paths`:
```python
('*/com.apple.MobileSMS/Previews/Attachments/*.ktx',)
```
Validated with `fnmatch.fnmatchcase` against real path shapes; `*` spans `/` so
the single pattern covers the `<uuid>/<uuid>/<uuid>/` nesting.

A file that is genuinely KTX1, or not a texture at all, is still listed with
whatever metadata parsed plus a `Status` note, but no image (parser returns
warnings instead of raising).

## Testing

- `pylint --disable=C,R`: 10.00/10
- Loads cleanly through `plugin_loader`; no duplicate artifact `name`.
- Run against the HC iPhone XS (iOS 18.7.8) extraction. Per-corpus row counts are
  not yet recorded in `sample_data` — will tighten the `sample_data` string to
  the `iOS <ver> | <n> rows` convention once counts are confirmed.

## Not done / follow-ups

- No coverage yet for older iOS generations or for confirming which `.ktx` under
  this path are ATX vs. genuine KTX1 across images.
- `sample_data` currently records the test image but not a row count.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/artifacts/apple_sms_preview_ktx_atx.py
+++ b/scripts/artifacts/apple_sms_preview_ktx_atx.py
@@ -1,0 +1,147 @@
+"""Apple Messages attachment previews stored as ATX textures with a .ktx extension."""
+
+from pathlib import Path
+
+from leapp_functions.parsers.apple_atx import decode_atx_file
+from scripts.ilapfuncs import (
+    artifact_processor,
+    check_in_media,
+    convert_unix_ts_to_utc,
+    logfunc,
+)
+
+__artifacts_v2__ = {
+    "apple_sms_preview_ktx_atx": {
+        "name": "Apple SMS Preview Attachments (ATX in KTX)",
+        "description": "Messages attachment preview thumbnails cached under com.apple.MobileSMS "
+                       "that carry a .ktx extension but are AAPL ATX texture containers, "
+                       "decoded to images when possible",
+        "author": "@charpy4n6, Claude",
+        "creation_date": "2026-09-09",
+        "last_update_date": "2026-09-09",
+        "requirements": "astc_decomp_faster, liblzfse",
+        "category": "SMS & iMessage",
+        "notes": "iOS caches preview thumbnails for Messages attachments under "
+                 "com.apple.MobileSMS/Previews/Attachments. Some of these files use a .ktx "
+                 "extension but the container is Apple's ATX format (AAPL magic followed by "
+                 "HEAD/astc/LZFS chunks), not Khronos KTX1. The ATX layout was derived from "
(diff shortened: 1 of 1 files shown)


abrignoni:
Pushed two commits to this branch while reviewing.

The first fixes the source path return, which was the only thing failing CI. It now returns the distinct `Previews/<subdir>` roots rather than one directory per attachment UUID, so the report cites one line per cache directory. It also corrects the description, which said every matched file is an ATX container, and fills in `sample_data` from real runs.

The second broadens the artifact. The glob only covered `.ktx` under `Previews/Attachments` and only decoded ATX, so ten files that are Khronos KTX1 produced a row with no image, and the sibling folders in the same cache were unreported. It now reads `.ktx`, `.jpeg` and `.png` under `Previews` and picks the decoder from the file's magic bytes rather than the extension, because the two disagree here. Files named `.ktx` are ATX, KTX1, or KTX1 behind an eight byte length, and some files named `.jpeg` are ATX. KTX1 goes to `scripts/ktx/ios_ktx2png`, the reader App Snapshots already uses. Two columns are added, Preview Type and Container.

Measured across 25 extractions: 122 rows becomes 174, and rows with a decoded image goes from 112 to 170. The four with no image are ATX with pixel format (1, 5), which the ATX parser does not support. `Previews/Search` turned out to hold 46 cached PNGs that are per-device content, and nothing was surfacing them.

The display name is now "Apple SMS Preview Cache", since the old one is wrong once PNG and JPEG rows are in the table. The artifact key and the module filename are unchanged.

Thanks for the artifact, the ATX in KTX catch is a good one.

PR #2166, opened 2026-09-12
Add Anonymous Chat & Fun iOS parser

darren33:
# Add Anonymous Chat & Fun iOS parser

Adds six artifacts for Anonymous Chat & Fun (`com.anonimchat.app`): application
information, accounts, conversations, messages, blocked users and media metadata.
Containers are identified from app metadata and iTunes domain names. No container
UUID is hardcoded, and SQLite databases are opened read-only.

Messages have a LAVA conversation view. Direction requires a signed-in username
from the same container's AsyncStorage manifest and agreement with the stored
sender name and flag. Missing or contradictory evidence leaves direction blank;
without a confirmed local endpoint, the sidebar lists the stored participants.
Account evidence cannot cross containers or extraction roots.

Attachments appear on messages only through a stored local path, an SDImageCache
key derived from the stored URL, or an explicit `message_info` join to a media
record. Photos originals use the app media ID and `ZASSET.ZUUID`; a message still
needs its own recorded join. Size and timestamps never establish a link. Bundle
resources are excluded, ambiguous matches stay unlinked, and remote URLs are
never fetched.

Media inventory covers folder, ZIP, TAR, hashed iTunes and raw-image inputs.
Extensionless cache files use stored URL keys and MIME metadata where available.
Classification and matching use indexes, and the analysis is shared between
Messages and Media while each artifact registers its own Media Manager references.

## Test data and validation

The six committed archives are subsets of eight SQLite/plist/JSON files generated
from literal invented values. They contain no casework or media payloads. The
provenance test compares JSON/plist bytes and SQLite schema/rows, avoiding SQLite
version-dependent headers; manifest JSON uses LF on every platform. The archive
comparison skips only when the runtime-contract sparse checkout omits case data.

Previously recorded examiner validation is included in `sample_data`: iOS 26.3;
Application Info 1; Accounts 2; Conversations 479; Messages 5,837; Blocked Users 6;
Media Files and Metadata 1,348. Those are aggregate counts only. No case records,
identifiers, paths, message content or media payloads are included.

Checked locally on Windows with Python 3.10:

- 61 focused tests pass, including the raw/iTunes cache failures, contradictory
  account evidence, source isolation and the extensionless-file performance case.
- The full admin suite completes 548 tests: 516 pass and 32 skip.
- All six strict fixture comparisons pass; pylint reports 10.00/10.
- Folder, ZIP, TAR and iTunes CLI tests generate reports from fresh synthetic
  metadata. Raw and iTunes media staging/registration tests use mocks and read no
  media bytes. No native macOS run or fresh case-media run is claimed here.

To reproduce from an iLEAPP checkout:

```text
python -m unittest discover -s admin/test/scripts -p "test_anonymous_chat*.py"
python admin/test/scripts/run_test_cases.py --module anonymousChat --strict
```

To rebuild the full synthetic input, choose a new directory:

```text
python admin/test/scripts/build_anonymous_chat_fixture.py --output-root scratch/anonymous-chat-new/extraction
```

## Interpretation limits

ZIP/TAR created and access times are blank. ZIP DOS timestamps are labelled with
their unknown timezone; folder timestamps describe the supplied copy. iTunes
times come from Manifest properties, and raw times from seeker metadata. Text
conversation times without a timezone remain in the raw field.

Raw and iTunes seekers do not expose a bounded header-read API. An extensionless
cache object without identifying metadata remains a `Filesystem Candidate` with
type `Unknown`; that does not establish that it contains media. Such a candidate
is not attached to a message without a recorded relationship. ZIP/TAR/folder
signature reads, when needed, are limited to the first 4096 bytes.

Filesystem rows represent associations, so one file may have several message rows;
count distinct filesystem paths to count files. Photos originals are labelled
because they can differ from the transmitted attachment. Standalone databases
without an identifiable app container are not targeted. The synthetic tests cover
metadata/report behavior; they do not establish association correctness in a case.


abrignoni (changes requested):
Thanks for this, @darren33, and for fixing the lint warnings so quickly! There's a lot of careful work here. Finding the containers from the app's own metadata instead of a hard-coded UUID, refusing a database in a container it can't identify, scoping conversation keys to their database, and opening everything read-only are all good stuff. The SDImageCache link from the stored URL is a nice find too.

I went through the PR at `ee28b2f9` and ran it against test trees built with your fixture builder, adding media files, an iTunes-style backup and a raw image. A few things need to change before we can merge.

### Needs changes

**1. Media linked to messages by size and time**

[`_inferred_media_matches`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L1393) attaches a local file to a message when the byte size matches, the type is compatible and the file time is within 300 seconds. [`_media_table_message_matches`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L1493) does the same with a one-second window, and it's the only thing tying a Photos original to a message. The best candidate goes into the Messages `Media` column, so the picture shows up on the message.

In a test tree I saved an unrelated PNG of the right size 90 seconds after message-3, and it became message-3's picture. Nothing in the data tied the two together.

For chat media, a file should only go on a message when the app recorded the link: a stored path, the SDImageCache key from the stored URL, a join key or a hash. When there's no recorded link, please leave `Media` empty and say so in the notes. A size and time match can't tell the real attachment from another file of the same size, and once it's on the message it reads as the attachment. Does anything in `messages` (maybe `message_info`) record which `media` row a message used? That would be the recorded link for the Photos originals.

**2. ZIP times read as UTC**

[`_zip_datetime`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L663) treats a ZIP entry's DOS date and time as UTC, but those fields hold the local time of the machine that wrote the zip. On one of the iOS full file system extractions I test with, the DOS time is 7 hours behind the UTC time in the entry's extended timestamp field on 1,997 of 2,000 sampled entries.

That makes the result depend on the input type. In my test, folder and tar input gave message-3 a red decoy saved 90 seconds after it. A zip of the same tree, written on a machine five hours behind UTC, gave it a blue decoy saved five hours after the message. If any time comparison stays, [`FileSeekerZip.decode_extended_timestamp`](https://github.com/abrignoni/iLEAPP/blob/9b55395960bf5d696bf968330c99c4f98e1f4337/scripts/search_files.py#L876) already reads the UTC value.

**3. iTunes backups return nothing**

The iTunes seeker [maps `AppDomain-<bundle id>` to `Containers/Data/Application/<bundle id>`](https://github.com/abrignoni/iLEAPP/blob/9b55395960bf5d696bf968330c99c4f98e1f4337/scripts/search_files.py#L47), but [`_container_location`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L341) needs a UUID there. I built a minimal unencrypted backup (Manifest.db) with the fixture's app files under `AppDomain-com.anonimchat.app`. `anonimchat.db` was staged, and all six artifacts reported "No data found" with no error. In that path shape the folder name after `Data/Application/` is the bundle id itself, which is enough to identify the container.

**4. Raw image input loses all media**

[`_iter_source_entries`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L683) reads seeker internals (`zip_file`, `tar_file`, `_all_files`, `single_file_abs_path`), and the raw image seeker added in #2165 has none of them. I ran `-t raw` on an APFS image of the same test tree. Messages parsed, but there were no filesystem media rows and no media links, including the valid SDImageCache one, and no error. `seeker.search()` works on every input type. If that's too slow, please say in the notes which input types get media.

**5. The fixture test fails on macOS**

`test_published_samples_are_exact_subsets_of_a_fresh_synthetic_build` fails for all six archives on macOS (Python 3.12, SQLite 3.43.1):

- `manifest.json` in the committed zips has CRLF line endings, because [`write_text`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/admin/test/scripts/build_anonymous_chat_fixture.py#L184) translates newlines on Windows. `write_bytes`, or `newline='\n'`, fixes that part.
- `anonimchat.db` and `Photos.sqlite` store the SQLite library version in the file header (bytes 96 to 99): 3.37.2 in the committed files, 3.43.1 here. That can't be pinned, so comparing the tables and rows would be steadier than comparing bytes.

CI doesn't catch it because the runtime-contract job [checks out without `admin/test/cases/data`](https://github.com/abrignoni/iLEAPP/blob/9b55395960bf5d696bf968330c99c4f98e1f4337/.github/workflows/python_runtime_contract.yml#L40), so the comparison is skipped there.

**6. Real data**

Everything in the PR is tested on synthetic data. If you can run it on an extraction that has the app, please record the OS version and row counts in `sample_data`. We don't need the data itself.

### Notes and descriptions

- The [Messages notes](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L116-L117) say "In the observed schema, from_username is the local conversation account and to_username is the remote participant." Every direction label depends on that, and the notes don't say what it was observed on. What establishes it? If the app's AsyncStorage `manifest.json` holds the signed-in username, that would settle it. It's in `paths`, but nothing reads it right now.
- I couldn't verify App Store ID 1483337394. Apple's lookup API returns no result for that ID or for `com.anonimchat.app` in the storefronts I tried, and the App Store URL returns 404. Where did the ID come from? If it can't be confirmed, I'd keep only the bundle ID, which comes from the extraction. The appInfo notes also say the app is "[listed by MWM](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L12)". Is MWM the developer, or a site that lists apps? If it's a directory, I'd drop it so nobody reads it as the publisher.
- The media notes say ZIP input "may" have blank created and access times. For zip and tar input they're always blank. For folder input they come from `os.stat` on the copy iLEAPP was pointed at, so they describe that copy, not the device. I'd say that in the notes, or leave those two columns out.
- `conversation_time` is an integer in the fixture, but [`_human_time`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L456) parses `YYYY-MM-DD HH:MM:SS` and marks it UTC. On the fixture the parse fails and the raw integer goes into the datetime column, so neither path is really tested. Which form does the app store? If it's text with no zone, it shouldn't be stamped as UTC.
- The PR body says media rendering "was checked in LAVA with the parser running against the case", and the [module comment](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L240-L243) says development validation "deliberately does not run that media path against case material". No need to share any data, but the two should say the same thing.

### Smaller things

- The media inventory includes the app bundle's own images. All 40 PNGs I placed inside the `.app` came out as rows. On one test extraction, app bundles hold a median of 37.5 image, video or audio files each (up to 3,725). Those ship with the app, so I'd leave the bundle container out of the inventory.
- App Group containers keep the group id in `MCMMetadataIdentifier` (77 of the 80 I sampled start with `group.`), so the [comparison with the bundle id](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L378) won't find this app's group, and the AppGroup pattern stages every group's metadata for nothing.
- The Photos original lookup ([L1613](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L1613), [L1641](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L1641)) is an exact match on paths starting `private/var/` or `var/`. The same test tree zipped under a `filesystem1/` root lost the Photos link that the `/private/` root produced, and 7 of the 17 extractions I have with DCIM files use a different root like that.
- The fixture puts the bundle container at [`private/var/mobile/Containers/Bundle/Application`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/admin/test/scripts/build_anonymous_chat_fixture.py#L27). The iOS extractions I test with have it at `private/var/containers/Bundle/Application`, which your hardening test already uses. The fixture's app database also gets a [`ZASSET` table](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/admin/test/scripts/build_anonymous_chat_fixture.py#L93). Does the app's database really have one, or is that left over from the Photos fixture?
- [`app_plists['bundle']`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L385) and [`app_plists['data']`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/scripts/artifacts/anonymousChat.py#L391) are set but never used.
- In the tests, the fallbacks to [`build_synthetic_fixture.py`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/admin/test/scripts/test_anonymous_chat_fixture.py#L13) and [`testdata/anonymousChat_synthetic_extraction.zip`](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/admin/test/scripts/test_anonymous_chat_fixture.py#L35) point at files that aren't in the repo, and [one test asserts the author string](https://github.com/abrignoni/iLEAPP/blob/ee28b2f9b9feb48cb231df52341d735c1ea68668/admin/test/scripts/test_anonymous_chat_hardening.py#L310), which will break if that field is ever reformatted. The README in `admin/test/cases/` would fit better in the module notes or this PR.
- Speed: the matching compares every media message with every candidate file, and it runs for both Messages and Media. On synthetic data it cost about 7 microseconds per pair, so 5,000 media messages against 10,000 cache files would take roughly 6 minutes per artifact. Indexing the files once by path, basename and cache key would avoid the pairwise scan.

Thanks again for putting this together. Happy to answer anything here, and once these are addressed I'll take another look.

darren33:
Thanks for going through this. I’ve pushed the fixes and added the OS version and row counts to `sample_data`. Tests and pylint are passing.

Latest commit is `7433c6f`. Could you take another look when you get a chance?

Thank you.

